PR #45, opened 2026-07-09
perf(eliashberg_dynamic): parallelize spatial FFTs via scipy.fft workers

k-yoshimi:
## Summary

**Stacked on #44** (which is stacked on #43); only the last commit is new. Retarget after the parents merge.

Routes the dynamic kernel's spatial forward/inverse FFTs through `scipy.fft` with a worker-thread count (`_spatial_ifftn` / `_spatial_fftn`), controlled by **`[eliashberg] fft_workers`** (default `-1` = all cores; set smaller when running several dynamic solves concurrently, e.g. matching `OMP_NUM_THREADS`).

- Falls back to `numpy.fft` when scipy is unavailable or `workers == 1`.
- No-op on the CuPy backend (#44): device FFTs already run on the GPU, so `workers` is ignored there.
- This is a port of the un-PR'd branch `perf/dynamic-eliashberg-speedup` (d66a326, measured on the bp-ICl2 production runs: ~4x matvecs combined with the vertex precompute at norb=2, Nk=1024, nmat=1024; a real 8.3 GPa point dropped 60 s -> 28 s end-to-end with unchanged lambda) onto the backend-dispatched kernel. The ported version measures 1.8x over the serial hoisted kernel on an 8-core machine.

## Test plan

- New `test_kernel_fft_workers_matches_serial`: parallel path matches serial to ~1e-14 (skips if scipy.fft is unavailable).
- Full suite: 636 passed, 5 skipped (cupy tests, machine without GPU).


k-yoshimi:
**AI review cycle (Codex + Antigravity) — result**

Round-1 Codex must_fix applied here: `fft_workers` now defaults to **1** (serial numpy, bit-compatible with previous releases); the scipy-parallel spatial FFT is opt-in (`-1` = all cores). This also avoids thread oversubscription when many solves run concurrently. Docs updated accordingly (the PR description's original "default -1" is superseded).

PR #47, opened 2026-07-09
feat(flex): GPU (CuPy) execution + parallel spatial FFTs

k-yoshimi:
## Summary

**Stacked on #45** (#43 <- #44 <- #45 <- this); only the last commit is new. Retarget after the parents merge.

Adds `[mode.param] gpu = true` for the FLEX solver: after the (host) H0 eigendecomposition and initial mu search, the H0 eigenpairs and the interaction move to the device once and the **whole SCF loop runs on the GPU** — dressed G (batched inverse), chi0q (FFT pair bubble), chi_s/chi_c (batched solve), V_eff (batched matmul), and the self-energy FFT convolution. Results return to host numpy at storage. Same warn-and-fall-back behavior as the dynamic Eliashberg `gpu` flag when CuPy or a CUDA device is missing.

- The chemical-potential search is the one CPU exception: its operator is non-Hermitian and CuPy has no general eigensolver, so `_matsubara_number_operator` brings M to the host (the search was already eigenvalue-based, one decomposition per iteration).
- The spatial-FFT helpers move from `eliashberg_dynamic` to `backend.py` and now also serve `RPA._calc_chi0q` and the FLEX self-energy, so **`[mode.param] fft_workers`** (default -1) parallelizes the CPU spatial FFTs exactly like `[eliashberg] fft_workers` (#45). Ignored on the GPU (cuFFT).
- The converted shared RPA methods (`_calc_green`, `_calc_chi0q`, `_solve_rpa`) dispatch on their input arrays — the RPA solver itself is unchanged on CPU; its own `gpu` entry comes as a follow-up PR.
- Docs: FLEX tutorial parameter tables (EN/JA).

## Measured (NVIDIA RTX 6000 Ada)

1-orbital Hubbard, 64x64 k-mesh, Nmat=1024, 3 SCF iterations: **22.5 s (CPU serial FFT) → 2.4 s (GPU) = 9.5x**, with max|Sigma_gpu − Sigma_cpu| = 4.4e-15. (`fft_workers` alone: 1.1x here — the 1-orbital case is matmul-dominated, not FFT-dominated.)

## Test plan

- `tests/test_flex_gpu.py` (new): gpu=true fallback-with-warning (monkeypatched missing CuPy, identical outputs), full-solve GPU-vs-CPU equivalence incl. host-numpy output types (skips without a CUDA device), fft_workers serial-vs-parallel equivalence.
- On clavius (RTX 6000 Ada): 48 GPU-relevant tests pass. Locally: 638 passed, 6 skipped.


k-yoshimi:
**AI review cycle (Codex + Antigravity) — result**

Fixes applied on this branch after the review:
- `[mode.param] fft_workers` defaults to **1** (opt-in parallel FFT, matching #45's updated default; the description's original "default -1" is superseded).
- FLEX no longer mutates `ham_info.ham_inter_q` to a device array (local device copy instead).
- `H0_eigenvalue`/`H0_eigenvector` are restored to host arrays before `solve()` returns (incl. `IterationMax=0`).
- The dressed mu-search bracket uses the host band energies from `_matsubara_number_operator`.
- New GPU test: calc_mu (filling-driven) GPU-vs-CPU equivalence + host-attribute leak assertions.

Verified: 641 tests locally, 54 GPU tests on RTX 6000 Ada.

PR #48, opened 2026-07-09
feat(rpa): GPU (CuPy) execution via [mode.param] gpu=true

k-yoshimi:
## Summary

**Stacked on #47** (#43 <- #44 <- #45 <- #47 <- this); only the last commit is new. Retarget after the parents merge.

Consumes the `[mode.param] gpu` flag (introduced in #47) in `RPA.solve`: the H0 eigenpairs move to the device after the (host) initial mu search, so the bare Green's function, the chi0q FFT pair bubble, the spin-inflation einsums, the batched chiq solve, and the transverse (ring+ladder) channel all run on the GPU. `chi0q`/`chiq`/`chiq_pm` are stored back in `green_info` as host numpy arrays; an externally supplied `chi0q_init` is transferred to the device instead of recomputed. Same warn-and-fall-back behavior when CuPy or a CUDA device is missing.

- `_calc_chi0q_transverse` / `_build_transverse_channel` become array-backend agnostic like the longitudinal path, and the transverse spatial FFTs now also honor `fft_workers` on the CPU.
- Docs: `gpu` and `fft_workers` entries in the RPA config chapter (EN/JA), covering both RPA and FLEX.

## Measured (NVIDIA RTX 6000 Ada)

1-orbital Hubbard, 64x64 k-mesh, Nmat=2048: **5.6 s (CPU) → 1.2 s (GPU) = 4.6x**, max|chiq_gpu − chiq_cpu| = 4.8e-13.

## Test plan

- `tests/test_rpa_gpu.py` (new): gpu=true fallback-with-warning, GPU-vs-CPU equivalence for chi0q/chiq incl. host-numpy output types, and the ring+ladder transverse channel (chiq_pm) equivalence. CuPy tests skip cleanly without a device.
- On clavius (RTX 6000 Ada): **51 GPU-relevant tests pass** (RPA + FLEX + dynamic Eliashberg + backend). Locally: 639 passed, 8 skipped.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #51, opened 2026-07-10
perf(flex): closed-form mu-search eigenvalues on the solve backend

k-yoshimi:
## Summary

After the GPU stack (#43–#48), FLEX runs with a **filling target** were dominated by the dressed chemical-potential search (87% of wall time at 64x64, Nmat=1024 on GPU): the mu-search operator M is non-Hermitian, so the eigendecomposition ran host LAPACK geev on millions of tiny matrices plus a device-to-host copy of M, and every Newton step evaluated N(mu) on the host.

- New `_eigvals_small`: for **nd_block <= 2** (single-orbital, and spin-reduced two-orbital models such as organic dimer models) the eigenvalues are evaluated in **closed form on M's own array backend** (1x1: the diagonal element; 2x2: trace/determinant formula). nd_block >= 3 falls back to host geev as before. The mu search only consumes the eigenvalues as an unordered set, so ordering is irrelevant.
- `_number_from_eigs` dispatches on lam's backend and returns plain floats — the whole search stays on the GPU end-to-end. The CPU path also gets faster (vectorized formula instead of batched geev: 4.3 s → 0.05 s per decomposition at the benchmark size).
- Docs: FLEX tutorial gpu row updated (EN/JA).

## Measured (RTX 6000 Ada, 1-orbital 64x64, Nmat=1024, 5 SCF iterations, filling-driven)

**23.2 s → 7.95 s (~3x)**; the eigenvalue step drops 580 ms → 0.5 ms per iteration. Fixed-mu runs are unaffected (3.6 s). GPU mu agrees with CPU to 1e-10.

## Not included (tried and rejected)

Warm-starting the search from the previous iteration's mu with a narrow bracket: with a large un-converged Sigma the truncated-Matsubara N(mu) is **not strictly monotone**, so a narrow bracket can select a *different root* than the wide band-range bracket and silently change the SCF trajectory (observed divergence at 64^2, Nmat=1024). A code comment records this so it is not re-attempted.

## Test plan

- New unit tests: closed-form 1x1/2x2 spectra match geev as multisets (1e-12), 3x3 geev fallback, plain-float returns, GPU-vs-CPU `_find_mu_dressed` equivalence (1e-10).
- Existing mu-update suite (dressed-number identities, derivative FD check, SCF particle-number conservation, charge-gap safety) passes unchanged on the new path.
- 652 tests locally; 81 GPU-relevant tests on clavius.


k-yoshimi:
**AI review (Codex) — must_fix found and fixed**

The 2x2 closed-form discriminant originally used `tr² − 4·det`, which subtracts two nearly equal O(ω²) terms for the Matsubara-shifted operator M = iω·I − H0 − Σ and loses the O(1) remainder to cancellation (measured 2.5e-11 absolute eigenvalue error vs LAPACK at ω=10³, growing linearly with ω). Fixed to the algebraically identical **shift-invariant form (a−d)² + 4bc**, which stays at machine precision; a regression test compares spectra at relative tolerance 5e-15 for ω up to 10⁵ (the naive form fails it from ω=10²). All tests green locally and on the GPU.

k-yoshimi:
**AI review cycle — converged**

Phase 2 (Antigravity, docs/workflow) reported no findings for this PR beyond what Phase 1 already fixed (the shift-invariant discriminant, see the previous comment). No further changes.

PR #55, opened 2026-07-10
feat(flex): IR-basis Matsubara axis for the FLEX SCF (matsubara_basis=ir, Stage 2)

k-yoshimi:
## Summary

Stage 2 of the Phase-0-reviewed design `docs/design/ir-matsubara.md`: the **FLEX SCF loop runs its whole frequency axis on sparse-ir nodes** — `[mode.param] matsubara_basis = "ir"` (default "uniform" unchanged; reduced/squashed schemes; general+ir is a fail-fast error). Requires the optional `sparse-ir` package.

- **chi0 computed natively on bosonic nodes** (G evaluated exactly on the bosonic τ nodes; anti-periodic reflection = index reversal on symmetric node sets; τ→iν is the physical integral) — the uniform-FFT **O(β/Nmat) δ(τ) constant and aliasing artifacts do not arise by construction**. Σ = V·G on the fermionic τ nodes. chiq/V_eff reuse the per-frequency machinery unchanged.
- **μ search simplified**: n(μ) = −Re Tr G(τ=β⁻) through the fermionic fit (closed-form node eigenvalues as evaluators, analytic derivative through the linear fit); the tail-subtraction machinery and `coeff_tail` are bypassed (ignored with INFO).
- `sigma_init` (uniform grid) migrates onto the nodes via a full-grid fit with the sweep-safe residual policy **`sigma_init_on_error = warn|abort|zero`** (default warn).
- Always-on **coefficient-decay diagnostic** per iteration (design Sec. 5 layer 1) warns on insufficient `ir_wmax`.
- **Outputs densified onto the uniform Nmat grid** (exact format compatibility — the Stage-1 dynamic-Eliashberg loader consumes them unchanged); only coefficient arrays leave the device, expansion by direct-layout host GEMMs.
- Composes with `gpu = true`, Anderson mixing, `fft_workers`, and Stage 1 (#54).

## Measured (clavius RTX 6000 Ada; T=0.05, 64², Nmat=4096 → 42 nodes = 98× axis compression; 30 SCF iterations)

| | uniform | ir | speedup |
|---|---|---|---|
| CPU | 1014 s | 18.9 s | **54×** |
| GPU | 65.4 s | 12.0 s | **5.4×** |

(The IR runs are dominated by the fixed ~10 s densified-output generation — the SCF itself is ~0.1 s/iteration; Stage 3's IR-native output will remove that floor.)

## Accuracy gates (4×4 Hubbard fixture, T=0.5)

[...]

PR #59, opened 2026-07-10
fix(eliashberg_dynamic): IR mode on realistic off-site-interaction FLEX data (closes #57)

### LINKED ISSUE #57: IR-basis dynamic Eliashberg (matsubara_basis=ir) gives wrong/erratic λ on static-dominated FLEX data
## Summary

The IR-basis dynamic-Eliashberg mode (`[eliashberg] matsubara_basis = "ir"`, PR #54) produces **wildly wrong and erratic leading eigenvalues** on real, static-fluctuation-dominated FLEX output, while the `uniform` path on the same input is correct. On β′-(BEDT-TTF)₂ICl₂ cRPA data (8.3 GPa, U+V, T=0.005), the uniform solver gives λ = 0.3599; the IR solver gives:

| `ir_wmax` | λ (IR) | vs uniform 0.3599 |
|---|---|---|
| auto (= 51.7) | **1.98e9** | garbage |
| 0.5 | 0.386 | +7% |
| 1.0 | **1.153** | wrong |
| 1.5 | 0.390 | +8% |
| 2.0 | 0.390 | +8% |
| 3.0 | **1.636** | wrong |
| 8.0 | 2.63 | wrong |

The result is non-monotonic in `ir_wmax` and never accurate — best case ~8% off, neighbouring `wmax` values diverge. Two independent root causes, both in `src/hwave/solver/eliashberg_dynamic.py`.

## Root cause 1 — `_ir_auto_wmax` misreads the `hr` data structure (→ 15× band overestimate)

`_ir_auto_wmax` assumes `hr` is `{R: (norb×norb) matrix}` and takes a Gershgorin row-sum:

```python
for _, mat in hr.items():
    m = np.abs(np.asarray(mat, dtype=complex))
    tot = m if tot is None else tot + m
band = float(tot.sum(axis=-1).max())
```

But the `hr` passed in (from `sc._read_interaction_files`) is a **flat dict `{((Rx,Ry,Rz),(i,j)): scalar}`**. So `mat` is a scalar, `tot` collapses to a 0-d array, and `band` becomes the **grand total of |t| over every (R, orbital-pair)** — not a band bound.

Measured on this model (norb=2, 100 transfer elements):
- `band` (as computed) = **15.76 eV**
- actual band width (diagonalizing H(k) on a 24×24 mesh) = **1.085 eV** (range [6.388, 7.474])
- → **15× overestimate**, so `ir_wmax = 3·(15.76 + U) = 51.7 eV`, `Λ = β·wmax = 200·51.7 ≈ 1.03e4` → an ill-conditioned basis and λ = 1.98e9.

## Root cause 2 — `drop_constant` discards a physically-significant static component

`_ir_compress(..., drop_constant=True)` (used for χ_s/χ_c) fits a frequency-independent column and **discards it** as the "O(β/Nmat) delta(τ) discretization artifact". Near a magnetic instability χ_s is **static-dominated** (its ν-independent part IS the physics), and the fit puts a large coefficient on the discarded column, removing physical weight. On this data the code's own guard fires:

```
IR compress chiq_s: the discarded frequency-independent component (2.308e+03) is
unusually large (>5% of the data scale 3.054e+00). ... may indicate an unexpected
constant in the input data
```

χ_s here is nearly flat in ν (q=0 diagonal changes by ~1e-4 from n=0 to the grid edge) with a large static value — exactly the regime that matters for superconductivity. This is why the error is erratic even at a sane `ir_wmax` (root cause 1 fixed alone does **not** make IR reliable here: `wmax=2.0`→0.39 but `wmax=3.0`→1.64).

## Reproduction

FLEX (uniform) output already on disk; run the dynamic solver twice:

```toml
# uniform (correct): lambda = 0.3599
[eliashberg]
  frequency = "dynamic"
  chi0q_mode = "flex"

# IR (broken): lambda = 1.98e9 at auto wmax
[eliashberg]
  frequency = "dynamic"
  chi0q_mode = "flex"
  matsubara_basis = "ir"
```

Model: β′-(ET)₂ICl₂ RESPACK cRPA, CellShape=[1,32,32], Nmat=1024, T=0.005, filling=0.75, singlet. (Happy to share the `chiq_s/chiq_c/green.npz` fixture.)

## Suggested fixes

1. **`_ir_auto_wmax`**: compute the band bound from the actual dispersion — e.g. `max|ε_k − μ|` from a diagonalized H(k) on the k-mesh (already built for the run), or at least handle the flat `{((R),(i,j)): scalar}` `hr` layout so the Gershgorin sum is per-orbital rather than a grand total. As-is it over-estimates by ~15× on realistic multi-hopping models.
2. **`drop_constant`**: it is unsafe when the susceptibility is legitimately static-dominated (the SC-relevant, near-critical regime). Options: verify the discarded constant is actually O(β/Nmat) before dropping (and error/keep it otherwise), make dropping opt-in, or subtract only the analytically-known tail. The existing `>5%` warning correctly flags the failure but the run still proceeds to a wrong λ.
3. Consider erroring (not warning) when the discarded constant exceeds the data scale, since the result is then unusable.

## Impact

IR mode is the intended path for low-T / large-Nmat runs, which are exactly the near-magnetic-instability conditions where root cause 2 bites. On this real material IR is currently unusable; the `uniform` path (now GPU- and FFT-worker-accelerated) remains correct. Filing so the IR Stage-2/3 work can account for static-dominated susceptibilities.

*(Diagnosis produced with Claude Code.)*

k-yoshimi:
## Summary

Fixes #57 (IR dynamic Eliashberg wrong/erratic λ on β′-(ET)₂ICl₂). Investigation on the actual material data found **three root causes** — the two from the issue (one reinterpreted) plus a deeper one the issue's symptoms pointed to:

### RC1 — `_ir_auto_wmax` misread `hr` (issue root cause 1, confirmed)
The Gershgorin-style estimate treated the flat `{((R),(i,j)): scalar}` dict as matrices and produced the grand total of |t| (15.8 — mostly the ~7 eV **on-site energies**, which shift but do not widen the band) instead of a band bound (true dispersion range 1.09). Λ ≈ 1e4 → ill-conditioned basis → λ ≈ 2e8. **Fix**: diagonalize H(k) on the run's k-mesh; `wmax = 3·(range + max U)`.

### RC3 — the instantaneous vertex part is out of any bosonic IR basis (new; the reason λ was wrong even at sane wmax)
The pairing vertex's frequency-independent bare term `0.5·(S+C)` is a **δ(τ)** in imaginary time. `_ir_vertex_to_rtau` fitted it into the bosonic basis, aliasing it into an uncontrolled smooth function. **Invisible on every existing fixture**: pure-CoulombIntra Kuroki has `S+C = U−U = 0` exactly; any off-site `CoulombInter` (UV/UVg/…) breaks the cancellation — which is precisely the issue's material. **Fix**: split `V_inst` off before the basis fit and handle its τ-integral analytically, `∫dτ e^{iωτ} V_inst δ(τ) F(τ) = V_inst·F(0)` with `F(0) = (1/β)Σ_ν F` evaluated through the fermionic basis (new `IRAxis.u_zero_plus`).

### RC2 — `drop_constant` guard (issue root cause 2, reinterpreted)
On this data χ_s actually decays by 97% across the grid; the huge "discarded constant" (2.3e3) at auto wmax was the **collinear split of the ill-conditioned basis** (an RC1 symptom), not physical static weight. The guard is still needed: `_ir_compress` now raises a hard, actionable `ValueError` (instead of warning and proceeding) when the fitted constant is comparable to the data scale — the regime where the split is meaningless or the static physics cannot be separated at this Nmat.

## Validation on the real material (8.3 GPa UVg, T=0.005, CellShape [1,32,32], Nmat=1024)

| config | λ before | λ after |
|---|---|---|
| uniform (reference) | 0.2212 | 0.2212 |
| IR, auto wmax | **1.8e8** | **0.2379** |
| IR, wmax=2 / 3 / 5 / 8 | 0.386 / 1.596 / 0.390* / wrong* | **0.2395 / 0.2392 / 0.2385 / 0.2369** |

*issue table values for its variant; 3.0 re-measured here as 1.596. After the fix λ is stable within ~1% across the wmax series and within ~7.6% of uniform — the finite-Nmat input-quality scale that shrinks systematically with the FLEX Nmat (pinned by the existing e2e test). Runtime also dropped 2712 s → ~200-550 s per solve.

## Measured non-finding (documented in the gate test)
The max-norm IR-vs-uniform kernel difference on FLEX data (0.2–0.7, non-monotonic in wmax) is **pre-existing and physics-safe**: it concentrates on the high-|n| nodes (1e-4 at |n|≤5, where λ lives) and is bounded by the input's out-of-basis artifact content. A new synthetic smooth-vertex gate proves the kernel algebra itself is exact (1e-6/1e-4 across both β regimes); the old wide-wmax gate branch (which sat in the now-rejected degenerate drop_constant regime) is replaced by it.

## Tests
+5: instantaneous-vertex kernel gate (β=2 and β=200), off-site-interaction end-to-end matvec gate (CoulombInter fixture — the class of model that was never covered), dispersion-vs-grand-total auto-wmax unit, nonartifact-constant compress error, synthetic smooth-vertex algebra gate. Full suite 681 passed; `unittest discover` OK.

Note: will conflict textually with #56 (feat/flex-ir-stage3) in `solve_dynamic`'s `use_ir` block — trivial to resolve whichever merges first; the `V_inst` split applies identically to Stage-3 IR-native inputs (it happens after vertex assembly, independent of input encoding).

https://claude.ai/code/session_01GY2dJV9aUKEdWSuXpxXUWw


k-yoshimi:
## ai-review-cycle (research-software-review) — result

**Phase 1 (Codex, code lens): must_fix = 0.** Addressed in 2fc026a:
- should_fix (real GPU bug): the `V_inst` subtraction ran after `Vs_q_w` had been moved to the device, mixing a CuPy array with the host-built `V_inst`. Unreachable on the pure-CoulombIntra GPU fixture (there `V_inst = 0` — the same cancellation that hid RC3 itself), so it got dedicated coverage: new `test_solve_dynamic_ir_gpu_matches_cpu_offsite` runs `gpu=true` with a nonzero instantaneous vertex — **verified on clavius CUDA, 17/17**.
- optional: the off-site kernel gate now loops both pairing channels; the triplet `0.5(C−S)` vanishes identically on the 1-orbital fixture (asserted and documented — the channel-agnostic split+analytic-add-back algebra is covered for arbitrary tensors by the const-vertex gate).

**Phase 2 (user/workflow lens): 2 must-fix-class findings, fixed in 4b4d329:**
- Tutorial (EN/JA) claimed the frequency-independent constant is always discarded; it now documents the hard error with the printed remedies, refreshed stale accuracy numbers (1% → ~1.5e-2 @ Nmat=128, dispersion-based auto wmax), and gained a **migration warning**: pre-fix dynamic-IR results on off-site-`CoulombInter` models are wrong and must be recomputed.
- Sharp catch: the compress-error remedy recommended FLEX `matsubara_basis="ir"`, which does not exist on this branch (it lives in unmerged #55). The message now names only remedies available here, and the test pins the remedy set. (Once #55/#56 merge, the FLEX-IR route can be added back to the message/docs.)

**Verification:** full local suite 681 passed / `unittest discover` OK / clavius CUDA 17/17 / incremental Codex re-review of the fix commits: zero findings. CI green (lint + Python 3.9–3.12).

k-yoshimi:
**Reconciled with PR #58** (the parallel issue-#57 fix merged yesterday evening) and with #55/#56:

- **RC1/RC2 duplicates dropped in favor of #58's merged versions** (μ/filling-aware spectral-half-range auto `ir_wmax`; `keep_constant`/`error_on_large_constant` + `ir_keep_static_chi`). The two superseded regression tests were removed — #58's `test_eliashberg_ir_wmax_static.py` covers them.
- **RC3 (the instantaneous-vertex fix) is this PR's surviving core** — #58 does not address it, and it is what makes λ wrong on off-site-V models even at sane `ir_wmax`. All RC3 gates/tests and the migration docs remain (docs updated to also name `ir_keep_static_chi` among the remedies).
- Kept BOTH wide-basis checks in the matvec gate: #58's real-data wide-wmax branch (escape hatch) and this PR's synthetic exactness gates.

**Final validation on β′-(ET)₂ICl₂ 8.3 GPa UVg (merged #55+#56+#58+this)**: IR auto-wmax λ = **0.2383** vs uniform 0.2212 (7.7%, the finite-Nmat input scale) — from 1.8×10⁸ before. Full suite 715 passed; CI green.

PR #82, opened 2026-07-26
Bond-resolved interaction channels for the linearized Eliashberg solver (opt-in)

k-yoshimi:
# Bond-resolved interaction channels for the linearized Eliashberg solver (opt-in)

Adds an opt-in `[eliashberg] bond_channels` mode that resolves the inter-site Coulomb
interaction `V` into bond channels `Δr ∈ {0, ±R, …}`, so the triplet eigenvalue
`λ_t` acquires the V-dependence of the Onari mechanism (Onari–Arita–Kuroki–Aoki,
cond-mat/0312314). Default (`bond_channels = false`) behaviour is unchanged.

## Why

H-wave already reproduces Onari's density-driven physics quantitatively (χ peaks,
singlet d-wave λ). The one quantity that did not follow was **λ_triplet's
V-dependence**: H-wave gave a nearly flat λ_t where Onari finds it rising. Cause:
the interaction was collapsed to a scalar `V(q) = Σ_R V_R e^{iq·R}` per orbital
pair, i.e. only the **Δr=0 (Hartree/direct)** channel was kept. The **Fock/exchange**
channel — which lives in the `Δr=±R` bond channels and drives odd-parity pairing —
was dropped. That is a scope limitation in the inter-site-`V` regime, not a bug:
for on-site `U` there is no bond structure to lose.

## What

- New module `src/hwave/solver/bond_channels.py`: `ResolvedInteractionSet` (single
  source of truth for bond topology, reversal-closed + shell-filtered), the
  bond-resolved bubble `χ̄(q;Δr,Δr')` (Onari Eq. 14), the bare spin/charge matrices
  and the **complete bare particle-particle (Cooper) vertex**
  `V^pp_η = Q_η L_η Q_η + Q_η(D+D†)Q_η`, the enlarged RPA-ladder dressing, and a
  two-part Eliashberg operator (fluctuation via O(B)-factorised FFT convolution +
  instantaneous low-rank term with the cross phase).
- `sc.py` wiring behind five raising guards, resource preflight, additive output
  provenance, and the **λ = λ^pp + λ^fl attribution** on the symmetrised kernel.
- New keys: `bond_channels`, `bond_green`, `bond_diagnostics`, `bond_max_shells`,
  `bond_memory_cap_gb`, `bond_precondition_atol/rtol/dense_limit`; new odd
  `init_gap` seeds `f_x`/`f_y`. Documented EN + JA.
- **`spectral_shift` now works in power iteration**, not just arnoldi (runs on
  `K + σI`, reports `λ − σ`). Needed because a repulsive-dominant kernel has a
  negative dominant eigenvalue, which flips the power iterate's sign every step so
  the convergence test can never be satisfied. Benefits existing non-bond users too.

## Result

Single-band square lattice, `U=4`, `n=0.7`, `T=0.02`, static, 16×16, per-V
FLEX green fixtures:

| V | 0.0 | 0.4 | 0.8 | 1.0 | 1.2 |
|---|---|---|---|---|---|
| λ_t | 0.0705 | 0.0925 | 0.1218 | 0.1645 | 0.3101 |

Monotone, Δλ = 0.240 (×4.4); reproduced on 32×32 (Δλ = 0.262, 9 % agreement).
Reproducible from the TOML entry point via `bond_green` (matches the direct-operator
reference to 2.2e-07).

## Scientific caveats — please read before citing these numbers

These are the results of controls run during review, stated so no reader
over-reads the claim:

1. **63 %, not 100 %, of Δλ is bond-specific.** Zeroing the `m≠0` (Fock/bond)
   blocks reproduces the scalar arm *exactly* at every V, so the bond-specific
   share is 0.1506/0.2396 = 63 %. The remaining 37 % is the collapsed
   `−½ W_c χ_c W_c` term the scalar path already carries. (The separate statement
   "λ^pp ≤ 0, so 100 % of the rise is λ^fl" is a *different* — pp/fl —
   decomposition and must not be read as the bond-specific share.)
2. **The whole odd-parity sector inflates**, 3.5–4.7× for sub-leading states vs
   4.4× for the leading one (in the scalar arm the 2nd state inflates only 1.20×).
   This is a broad odd-channel enhancement, not a selective f-wave instability.
3. **f-character drifts**: the tracked state captures only ~21 % of the f form
   factor at V=0, rising to ~68 % at V=1.2. The accurate description is "an odd,
   non-p state that acquires f character as V grows". The p-content ≤ 0.003 bound
   is solid throughout.
4. **Static, not dynamic, and not conserving-FLEX.** Absolute λ is not comparable
   to dynamic/FLEX references (static/dynamic ≈ 2.3× near the instability). Onari
   Fig. 3 is fully dynamic, so **no quantitative literature comparison is claimed
   here**; that needs the dynamic port (+ optionally bond-resolved FLEX Σ).
5. **Proximity to the CDW**: V=1.2 sits ~85 % of the way to the extrapolated CDW
   at V≈1.44. The enhancement is a factor ~3, not a divergence (σ_min stays
   0.12–0.29, and the top odd λ = 0.31 is far from 1). A fixed-distance-from-CDW
   control has not been run.
6. **The scalar baseline is not flat** (126 % spread), contrary to the design's
   ≤5 % expectation — the recorded "flat" observation was the *dynamic* path. The
   acceptance test therefore asserts the discriminating statement instead:
   bond > scalar at every V>0 with a monotone gap reaching +0.150 (16×16) /
   +0.179 (32×32). Verified discriminating by mutation testing (sign-flipping the
   bond Fock blocks → negative gap; sign-flipping `V^pp_t` → non-monotone; zeroing
   the Fock blocks → gap ≡ 0).

## Pre-existing approximations this surfaces

- **The existing multi-orbital path collapses the inter-site Fock term.**
  `_build_sc_matrices_all_q` puts the whole Fourier sum `V_ab(q)` into the local
  `Δr=0` Fock element. The bond path corrects this, so a multi-orbital user with
  `R≠0` `CoulombInter` who enables `bond_channels` will get **different (more
  correct) numbers** than the scalar path. Multi-orbital is gated off at the top
  level pending its own validation, but this should be an issue.
- **At `norb=1` the general (4-index) vertex path drops `V` entirely** (Cases 2/3
  require `a≠b`), so its λ is exactly V-independent. The meaningful scalar
  comparison is the 2-index `_compute_vertices_simple` path (`Wc = U + 2V(q)`),
  which is what the acceptance test uses.

## Migration notes

- `spectral_shift` with `solver_mode="both"`: the power-iteration arm is now
  shifted too, so its reported number changes from an unsigned iterate norm to the
  signed eigenvalue. Labelled in `eigenvalue.dat`.
- Same for the dynamic iteration path, which previously ignored `spectral_shift`
  entirely; a matching note is now written to the dynamic output.

## Tests

917 → **1023 passed / 25 skipped**, zero regressions. Every physics component is
pinned by an *independent* oracle (literal direct summation for the bubble and the
`O(N_k²)` kernel; hand-computed analytic fixtures for the vertices), and the phase
conventions were verified discriminating by sign-flip mutation. `bond_channels=false`
is asserted bit-identical on existing fixtures. The L=32 grid-convergence test
regenerates its fixtures on demand (`pytest -m slow`), keeping the committed
fixture set at 1.54 MB.

## Known follow-ups (not blocking, opt-in path is guarded)

- Dynamic bond path + bond-resolved FLEX Σ (for the quantitative Onari Fig. 3
  comparison). Requires rethinking the Hermitian preconditions and the `√W`
  attribution metric, since the dynamic kernel is non-Hermitian.
- Multi-orbital validation (equations implemented, numerics gated off; the
  asymmetric-`V_ab ≠ V_ba` convention is now pinned by a unit test).
- Unrelated pre-existing issues found on the way, worth their own tickets:
  `_calc_chi0q_internal` is run-to-run non-deterministic at ~1e-17 (in `rpa.py`;
  the bond path never calls it), and **`pytest.ini` uses a `[tool:pytest]` header
  so none of its `addopts`/`markers` apply — the `--cov-fail-under=80` gate is not
  actually being enforced.**

https://claude.ai/code/session_01UuMPNJwk4oRsWF5MVZVqiV


k-yoshimi:
Before merging: the interaction-vertex content of the bond channels (S_bond/C_bond and V^pp) should be adjudicated against exact diagonalization at first order — the validation this PR carries (k-sum oracles, analytic fixture, scalar-arm consistency, literature curve) compares at the same approximation level and has historically missed exactly this class of error (#90, #104, #113). Tracking issue with the method: see the newly filed ED-adjudication issue. Also note: the test modules were converted to unittest.TestCase on this branch so the gating CI actually runs them (previously it collected zero tests from the five bond/spectral modules).

PR #87, opened 2026-07-27
Fix the orbital-pair embedding of reduced-scheme susceptibilities in the Eliashberg vertex

k-yoshimi:
## Problem

A `calc_scheme="reduced"` / `"squashed"` run stores its susceptibility in the reduced spin-orbital space, where the matrix index is a **density pair**: the stored `X[a,b]` *is* `chi_{(a,a),(b,b)}`. (Its companion interaction reduction is `einsum('ksasatbtb->ksatb', ...)` — the density-density diagonal of the vertex.)

Two places lifted that matrix into the `norb^2` orbital-pair space by reading the density-pair index as an ordinary orbital index:

```
out[(l1,l2),(l3,l2)] = X[l1,l3]        # a delta_{l2,l4} scatter, i.e. kron(X, I_norb)
```

That drops the inter-orbital density coupling `chi_{(0,0),(1,1)}` — a component `S @ chi @ S` genuinely references — and scatters `X` onto pair indices the reduced scheme never computed. The faithful embedding is

```
out[(a,a),(b,b)] = X[a,b]              # every other component zero
```

Affected sites:

1. `_expand_flex_chi` — the `chi0q_mode="flex"` route, **static and dynamic**. The dynamic kernel reads chi through the same loader, so the error entered at every frequency, where no alternative route exists to cross-check against.
2. `_compute_vertices_general`, 2-index branch — the `chi0q_mode="load"` route. Reached by the ordinary two-step workflow: RPA with `calc_scheme="auto"` resolves to `"reduced"` for U + Hund/Ising/PairHop, and `hwave_sc` then selects the general S/C formulation because an inter-orbital term is present.

Reported with a self-contained reproduction; confirmed on this branch's base.

## Symptom

Two-orbital Hubbard (U=8, 8x8, T=1, singlet, static), FLEX at `Mix=0`/`IterationMax=1` so Sigma=0 and the physics is identical to RPA:

| route | before | after |
|---|---|---|
| general RPA + `load` | 0.2269804250 | 0.2269804250 |
| general FLEX + `flex` | 0.2269804250 | 0.2269804250 |
| reduced RPA + `load` | 0.2269804250 | 0.2269804250 |
| reduced FLEX + `flex` | **0.2355141950** | **0.2269804250** |

## Verification

On a real two-orbital model at Sigma=0, every reduced (kuroki) result now equals its general (myo) counterpart to all printed digits, across the full frequency/axis matrix:

| | general (myo) | reduced, before | reduced, after |
|---|---|---|---|
| static / uniform | 4.90887226 | 2.84261111 | **4.90887226** |
| static / IR | 63.6140528 | 38.7024073 | **63.6140528** |
| dynamic / IR | 22.9817667 | 12.5716760 | **22.9817667** |
| dynamic / uniform | 3.14628784 | 3.02925673 | **3.14628784** |

`calc_scheme="squashed"` stores a byte-identical layout to `"reduced"` and reproduces the general lambda exactly, so it is covered by the same fix.

The second site is pinned by feeding a chi0q that is *exactly* density-diagonal, which makes the 2-index reduction lossless — so the 4-index and 2-index vertices must then agree exactly. They previously differed by up to 5.7 in absolute value.

GPU was not exercised (no CUDA device available), but it cannot diverge: the expansion is a single shared host-side implementation, run before anything moves to the device on the static path and delegated to by the dynamic path. A test pins that it returns a host array.

## Scope of the behaviour change

Results change for `norb >= 2` runs whose susceptibility came from a `reduced`/`squashed` scheme — static and dynamic — whenever the inter-orbital density components are nonzero, which is the generic case.

- `norb = 1` is bit-identical, asserted with `np.array_equal` against the previous placement.
- General (`myo`) results are untouched; `ring+ladder` output is bit-identical for spin-free and spin-diag.
- With `CoulombIntra` only, a reduced run now reproduces the equivalent `load` and general-scheme results exactly.

The previous release note scoped this class of change to `norb = 2`; that was too narrow, since for `norb >= 3` the spin-orbital block was extracted correctly but still scattered as `kron(X, I_norb)`. Corrected in the docs (EN+JA), together with an explicit directive to recompute stored eigenvalues and gap functions.

## Also in this PR

Found while auditing for defects of the same class.

**A spin-polarized reduced susceptibility was consumed silently.** The embedding keeps only the spin-up block, which is exact only when the discarded blocks are redundant. `spin_mode` is auto-detected, so an ordinary input with `coeff_extern` plus an `Extern` field produces genuinely different blocks — reproduced on a 2-orbital model where they differ by 23.7 against a block scale of 14.9, and `hwave_sc` returned an eigenvalue with no indication at all. The solver now inspects the stored data (the npz records no `spin_mode`) and reports non-redundant discarded content, naming the channel and the magnitude. The check is report-only; the returned tensor is unchanged, pinned by a test.

**Two dead reduced expansions in `rpa.py`'s transverse channel are removed.** They carried the same defective placement but were unreachable — `calc_type="ring+ladder"` is the only caller and `sys.exit()`s unless `calc_scheme="general"`. Verified by making both branches raise: the suite passes, and reduced/squashed inputs are refused by the guard while auto/general take the general path. They were left with a comment asserting the delta structure is correct, which would have let a future relaxation of that guard silently reintroduce the bug. The spin-free branch keeps an explicit shape assertion naming the invariant and recording what a reduced chi0q would actually require.

**A warning for interactions the reduced data cannot dress.** `CoulombInter`, `Hund`, `Ising`, `Exchange` and `PairHop` put weight on the off-density S/C blocks where a reduced run computed no susceptibility, so those channels enter the vertex undressed. The decision is made on the *assembled* S/C matrices rather than per term, because those blocks are sums and the terms can cancel there (`Exchange = -PairHop` is covered by a test). Emitted once per run, not once per Matsubara frequency.

## Audit notes

- No spectator (`delta_{l2,l4}`) scatter remains anywhere in `src`.
- Every bit-identity the code asserts about its own scatters was checked numerically and holds, including the premise of this fix: the reduced `chi0[a,b]` really is the general `chi0[a,a,b,b]`.
- Sublattice folding (`SubShape != [1,1,1]`) writes chi with the folded `norb` while `hwave_sc` uses the unfolded one; the shape guard rejects it with a clear `ValueError` on both paths, so it fails loudly rather than silently. Not addressed here.

## Tests

New `tests/test_flex_reduced_eliashberg.py` covers the embedding, the physical equivalence to the simple/load vertex, the density-diagonal invariant on the load route, the `norb=1` bit-identity, the discarded-spin detector, and both warnings — including that each fires exactly once through the real public entry points. `tests/test_eliashberg_dynamic.py` gains a dynamic-loader check across every frequency.

Four existing tests encoded the old placement or overclaimed, and are corrected in place with a note on why:

- `test_vertex_4index_vs_2index_differ` asserted the 4-index and 2-index vertices disagree for a `CoulombIntra`-only model. With S and C confined to the density block there, the reduced chi0q carries everything the vertex reads, so they must now agree exactly; the test asserts that, and separately that they still differ once `Hund` reaches the off-density blocks.
- `test_rpa_chi_s_manual_reference` built its reference from `chi0 = 0.05*I`, whose off-density entries are not representable in the 2-index chi0q it then handed the solver. (The old scatter reproduced that identity by coincidence, since `kron(0.05*I2, I2) == 0.05*I4`.) It now uses a density-diagonal chi0.
- `tests/test_flex.py`'s `_flex_chi_to_eliashberg` mirrored the old layout. Its callers are single-orbital, where that coincides with the correct placement, so it survived the fix silently and would have validated the wrong layout for any multi-orbital caller. It now delegates to the production expansion.
- `test_rpa_chi_vs_rpa_solver` claimed in its name and docstring to compare `sc.py`'s S/C dressing against the RPA solver, but only asserted finiteness — it never touched the solver's chi. Renamed to what it does, with the gap recorded. A real cross-implementation comparison is worth having, but the reduced RPA solve and `sc.py`'s general S/C formulation do not treat `Hund` identically, so the quantity they should agree on has to be established rather than assumed.

935 passed. Documentation builds at the same warning count as `develop` in both languages.

## Open question

Both new conditions are warnings rather than hard errors, so existing runs keep working. Say the word if a hard failure is preferred for either — particularly the spin-polarized one, which is undefined behaviour rather than a documented approximation.


k-yoshimi:
Superseded by PR #121, now merged into develop. It lands the same density-pair embedding fix rebased onto the current code (this branch had conflicted), and verified the bug was still live before fixing: the reduced two-index susceptibility X[a,b] is chi_{(aa),(bb)} and is now lifted onto exactly the density-pair slots, where the old expansion scattered it across same-orbital-offset slots (dropping the inter-orbital density correlations and fabricating dressing). PR #121 additionally validates the dressed Green's function file at the load boundary and rejects reduced susceptibilities combined with Exchange/PairHop, whose vertex content a reduced run cannot represent.

Issue #88, opened 2026-07-27
Spin-polarized Eliashberg: transverse channel in FLEX and an S_z-resolved pairing vertex


k-yoshimi:
Wish-list item, recorded from the discussion on #87. **Not urgent, and best attempted after the bond-channel work has settled**, since that touches the same vertex construction.

## What is missing today

The Eliashberg pairing vertex is paramagnetic: the Kuroki `S`/`C` matrices are `norb^2`-sized with no spin index, and the singlet/triplet decomposition assumes spin-rotational symmetry. As of #87 a spin-resolved reduced susceptibility is refused rather than silently reduced to its spin-up block, so the current behaviour is honest but limited:

| model | today |
|---|---|
| paramagnetic | supported |
| spin-polarized, no spin mixing (`Extern` + `coeff_extern`) | refused at the loader |
| spin-mixing / spin-orbit (`enable_spin_orbital`) | refused on a shape mismatch |

## Why the paramagnetic case needs nothing extra

For a paramagnetic system the transverse component is already fully contained in the spin channel. Measured on a two-orbital on-site-U model with `calc_type = "ring+ladder"`:

```
max|chi_pm|              = 7.136267278890249
max|chi_zz| (= uu - ud)  = 7.136267278890251
max|chi_pm - chi_zz|     = 1.8e-15
```

They agree to round-off, as SU(2) symmetry requires. The `3/2 * chi_s` factor in the FLEX effective interaction (`_calc_veff`) is exactly the count of three spin components — one longitudinal, two transverse — so a paramagnetic FLEX self-energy already carries them.

## Why the polarized case is not a post-processing step

Once the spins are split, `chi^{+-} != chi^{zz}` and the single `3/2 * chi_s` factor no longer collects them. The transverse channel then enters the self-consistency:

```
Sigma_up  <-  V_eff(chi^{+-}, chi^{zz}, chi_c)  <-  chi^{+-}  <-  G_up, G_down  <-  Sigma_up
```

so it cannot be computed once after convergence. (In RPA it can, because RPA has no self-consistency — `calc_type = "ring+ladder"` there produces `chiq_pm` as an extra output and leaves `chiq` untouched.)

## Rough scope

[...]

PR #89, opened 2026-07-27
Add an undocumented output_chi_pm switch to cross-check the FLEX spin channel

k-yoshimi:
Small verification-only addition, independent of #87.

## What it does

FLEX solves the **longitudinal** spin channel from the Kuroki `S` matrix. With `output_chi_pm = true` in `[mode.param]` it additionally solves the **transverse** (spin-flip) channel from the crossed Fock-exchange Hartree vertex (`RPA._build_transverse_channel`) and saves it as `chiq_pm`.

For a paramagnetic system SU(2) symmetry forces `chi^{+-} == chi^{zz}`. The two are assembled from different vertices, so their agreement checks the vertex construction rather than restating one computation. On a converged two-orbital on-site-U run:

```
max|chi_s|          = 1.501079486128
max|chi_pm - chi_s| = 1.7e-16     (relative 1.1e-16)
```

## Why it is undocumented

It produces a diagnostic output, not physics, and is redundant by symmetry — for any paramagnetic run the answer is already in `chiq_s`. It is off by default, and nothing consumes `chiq_pm`; the Eliashberg loader never reads it. Adding it to the manual would invite users to turn on a redundant computation.

## Where it does not apply

Both cases warn with the reason instead of silently doing nothing:

- **reduced / squashed** — the transverse channel needs the rank-4 orbital `chi0q` that only `calc_scheme = "general"` produces. There is also nothing to check on that path: `chi_s` already *is* the isotropic spin susceptibility.
- **spin-polarized** — `chi^{+-} == chi^{zz}` does not hold, so the comparison would be meaningless. Computing `chi^{+-}` for a polarized run is a much larger change: it enters the FLEX self-consistency (`Sigma <- V_eff(chi^{+-}, ...) <- chi^{+-} <- G <- Sigma`) rather than sitting on top of a converged result. Tracked as #88.

## Tests

`tests/test_flex_chi_pm_check.py`: off by default produces nothing; the transverse and longitudinal channels agree for a paramagnetic run; the reduced path skips with an explanatory warning.

908 passed.

### DIFF
--- a/src/hwave/solver/flex.py
+++ b/src/hwave/solver/flex.py
@@ -263,17 +263,27 @@ class FLEX(RPA):
                 .format(self.mixing_scheme))
         self.anderson_depth = int(self.param_mod.get("anderson_depth", 5))
 
-        # UNDOCUMENTED verification switch, off by default. On the paramagnetic
-        # general path it additionally solves the TRANSVERSE (spin-flip) channel
-        # and stores it as chiq_pm. For a paramagnetic system SU(2) symmetry
-        # forces chi^{+-} == chi^{zz}, and the two are built along genuinely
-        # different routes -- the longitudinal one from the Kuroki S matrix, the
-        # transverse one from the crossed (Fock-exchange) Hartree vertex -- so
-        # their agreement is a real cross-check of the vertex construction
-        # rather than a restatement. It is deliberately not in the manual: it
-        # adds a diagnostic output, not physics, and is redundant by symmetry.
-        self.output_chi_pm = _bk.as_bool(
-            self.param_mod.get("output_chi_pm", False))
+        # UNDOCUMENTED verification switch, off by default. On the general path
+        # it additionally solves the TRANSVERSE (spin-flip) channel from the
+        # crossed (Fock-exchange) Hartree vertex, which for an SU(2)-invariant
+        # interaction must reproduce the longitudinal spin channel built from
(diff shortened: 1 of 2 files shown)


k-yoshimi:
Closing: the cross-check this adds does not actually hold, and I would rather withdraw it than ship a verification tool that reports false results.

Review raised that the comparison could only be valid for `CoulombIntra`, where the longitudinal and transverse effective vertices are the same diagonal matrix and any leg-convention mistake is invisible. Measured, on a two-orbital model:

| interaction | relative `max|chi_pm - chi_s|` |
|---|---|
| `CoulombIntra` only | 2.3e-16 |
| `CoulombIntra` + on-site `CoulombInter` | **1.13** |

On-site `CoulombInter` is a density-density term and manifestly SU(2)-invariant, so `chi^{+-}` and `chi^{zz}` must agree — they disagree by 100%. The 1e-16 agreement I reported earlier was therefore the trivial case, not evidence that the construction is right.

The cause is a leg-convention mismatch, as review identified: the longitudinal channel is solved in MYO pair order and transposed on output, while the transverse channel is solved directly in the public `[a,c,b,d]` order. Fixing it means re-deriving the transverse vertex in the same orbital-leg convention as the quantity it is compared against — a derivation, not a switch to add — and I am not confident enough in doing that correctly to leave the feature open in the meantime.

Recorded in #88, which already tracks the transverse channel, together with the measurement above so the next attempt starts from a known failing case rather than the passing one.

PR #98, opened 2026-07-28
Build the interaction vertex in the orbital orientation the multi-site RPA requires

### LINKED ISSUE #96: sc.py builds the interaction vertex without the pair-index transpose the RPA and pairing vertices require
**Priority: high, but the direction has to be settled before anything is changed.** Unlike #95 this is not a missing term — both sides are self-consistent, they just disagree about which orbital index is which, so "fix it" means choosing which one defines the model.

## Symptom

Found alongside #95, from the same verification run (`Sigma = 0` FLEX should reproduce RPA-Eliashberg). After #95 is corrected the two paths still differ; aligning the orbital order as well makes `chi0q_mode = "load"` and `"flex"` agree to 9 digits (0.915193164), and the `Sigma = 0` FLEX run reproduces RPA-Eliashberg exactly.

## The mismatch

The susceptibility side and the pairing-vertex side pack the same interaction file into transposed orbital order.

`rpa.py`, `Interaction._make_ham_inter` — for an entry `(irvec, (a, b))` it writes the value into the **swapped** orbital slot at the same `irvec`:

```python
# beta beta' alpha alpha'
orb = (s4, b, s3, b, s1, a, s2, a)
ham_r[(*irvec, *orb)] += v * w
```

`sc.py`, `_build_interaction_k._to_k` — writes it straight:

```python
val_k[orb1, orb2, ...] += value * np.exp(-1j * (kx*Rx + ky*Ry + kz*Rz))
```

So `V_rpa(q) = V_sc(q)^T`. For a Hermitian-closed real interaction `V(R)^T = V(-R)`, hence `V_rpa(q) = V_sc(-q)`: the two coincide only where `q ≡ -q`, which is why high-symmetry points hide it.

**This is not an input-validation problem.** It persists for a perfectly and consistently declared interaction; #93 (unvalidated orbital symmetry of interaction files) is adjacent but distinct.

## Note that `rpa.py` is internally inconsistent too

In the same file, the *transfer* term is packed with no swap:

```python
tab_r[(*irvec, *orbvec)] += v      # _make_ham_trans
```

so within `rpa.py` the transfer uses `[a,b]` at `R` while the interaction uses `[b,…,a,…]` at the same `R`. That asymmetry is the structural origin of the mismatch and is worth resolving regardless of which convention is chosen.

## Relation to `db401fd`

`db401fd` ("align hwave_sc k-space builders with the solver-core convention", in #79) already aligned `_build_hamiltonian_k` and `_build_interaction_k` to `M[a,b](k) = sum_R M_R[a,b] e^{-ikR}`, i.e. to `_make_ham_trans`. `_make_ham_inter` was not part of that change, and it is the remaining outlier. So this issue is what is left after `db401fd`, not a regression from it.

## When it bites

Only when the interaction is genuinely asymmetric between `R` and `-R` in its inter-orbital components — then `V(q)` acquires complex off-diagonals and the two orders describe different models (mirror images of the bond arrangement). β-(meso-DMBEDT-TTF)₂PF₆ is such a case. Models with real, `R`-symmetric interactions are unaffected element-wise, which is why this survived.

## What has to happen first

Both conventions are internally consistent, so this cannot be settled by reading the code — picking the one that "looks right" risks re-labelling a model that has already been validated against reference data. The lesson from #91 applies directly: there, the codebase's own "independent ground truth" shared the convention it was validating and confirmed the wrong answer.

Settle it against something outside both paths. Concretely: build a small model with a deliberately asymmetric inter-site inter-orbital bond, evaluate the susceptibility directly from the declared bonds (or by exact diagonalization of a small cluster), and see which packing reproduces it. That fixes the susceptibility side; the pairing side then has to match it, not the other way round.

Only after that should either `_make_ham_inter` or `_build_interaction_k` be changed — and if `rpa.py` turns out to be the one that is wrong, the RPA reference data has to be re-examined too.

## Suggested tests

- The cross-path check that found this: `Sigma = 0` FLEX must reproduce the RPA-Eliashberg eigenvalue, on a model whose interaction is asymmetric under `R -> -R` so `V(q)` is genuinely complex off-diagonal.
- A direct assertion that `_make_ham_inter` and `_build_interaction_k` produce the same `V(q)` for the same input, away from high-symmetry `q` — the check that would have caught it at the unit level.

Related: #95 (the missing charge-diagonal term; both must be fixed for `load` and `flex` to agree), #93 (interaction files are not checked for orbital symmetry).

k-yoshimi:
`Interaction._make_ham_inter` placed an input entry `(R, (a, b))` into the orbital slot `(b, b, a, a)`, so after the `e^{-iqR}` FFT the reduced vertex was `V(q)` **transposed** in the orbital indices. `sc.py::_build_interaction_k` builds `V(q)` untransposed. The susceptibility was therefore solved with one orientation and the pairing vertex assembled from the other.

Closes #96

For a real, `R`-symmetric interaction the two coincide element-wise, which is why this survived. For a bond geometry that is not symmetric under `a ↔ b`, `V(q)` is complex and the two describe mirror-image models.

## Which orientation is correct

The sources cited elsewhere in this codebase cannot settle it. Mochizuki–Yanase–Ogata (cond-mat/0407094) and Kuroki–Aoki (0902.3691) both treat **on-site interactions only**, and their vertex matrices are symmetric in the orbital pair — MYO Eq.(6) gives `[U^c_1]_{aa,bb}` = `U` for `a = b` and `2U′ − J_H` for `a ≠ b` — so the orientation is invisible in their scope. H-wave's q-dependent inter-site `V` is an extension beyond them.

The multi-site extended-Hubbard RPA that H-wave generalizes does fix it: **Y. Mizuno, A. Kobayashi and Y. Suzumura, "Role of charge fluctuation in Q1D organic superconductor (TMTSF)₂ClO₄", [arXiv:1010.1084](https://arxiv.org/abs/1010.1084)** — a four-site unit cell with anisotropic intersite repulsion, i.e. exactly the case where `V_ab(q)` is complex.

| | |
|---|---|
| Eq. (4) | `V_αβ(q) a†_{k+qσα} a†_{k′−qσ′β} a_{k′σ′β} a_{kσα}` → **α carries +q** |
| Eq. (6) | `V_13(q) = V_13 + V′_13(e^{iq_x}+e^{−iq_y}) + V″_13 e^{i(q_x−q_y)}`, `V_βα = V*_αβ` → complex, Hermitian but **not symmetric** |
| Eq. (13) | `(χ̂⁰(q))_αβ = −(1/N_L) Σ …` = `−(T/N) Σ_k G_αβ(k+q) G_βα(k)` → **the chi0 convention this code already uses** |
| Eq. (12) | `χ̂^C = (Î + 2χ̂⁰V̂ + χ̂⁰Û)^{-1} χ̂⁰`, "the hat denotes a matrix representation with an element `(f̂)_αβ = f_αβ`" → **V̂ untransposed** |

Independently, with no reference to any convention in this codebase: for a density-density interaction the exact interaction energy is `Σ_q Σ_ab V_ab(q) ρ_a(q) ρ_b(−q)` with `V_ab(q) = Σ_R V_ab(R) e^{−iqR}`. Checked against the real-space `Σ_{i,R,a,b} V_ab(R) n_{ia} n_{i+R,b}` for a classical density configuration on a bond set with `V_ab(R) ≠ V_ba(R)`: that pairing reproduces it to 2.7e-15, the transposed pairing is off by 16%.

## The 2-orbital RPA reference is corrected too

`tests/test_rpa_2orb.py` compares against `RPATwoOrbital`, a hand-written helper **in the same file**. It cannot arbitrate — git history shows it was adjusted to agree with the solver twice:

- `4d60115` ("fix fft direction of interaction term") changed the solver's interaction FFT from `ifftn*nvol` to `fftn` and, in the same commit, flipped the helper's `Vxq` sign to match. The old line is still there, commented out.
- `edcf5cd` ("tests updated") **mirrored the fixture's bond directions** in `coulombinter.dat` and, in the same commit, added `.T` to the helper's RPA solve and a compensating `np.transpose(y, (0,1,3,2))` to the hamiltonian comparison — instead of updating `Vxq` for the new geometry.

`_make_ham_inter`'s placement itself has never changed since `5d4b5f8`. Verified both ways: with the pre-`edcf5cd` fixture and an untransposed helper the old solver passes and the corrected one fails; with the current fixture the opposite holds. The helper tracks whichever fixture it is paired with. Both compensating transposes are removed here, and the helper's `W` was already written in the paper's orientation (`W[0,3] = conj(Vxq) = V_01(q)`) — only its use was transposed.

## Tests

`tests/test_interaction_orientation.py` asserts both builders produce the same `V(q)` on a deliberately asymmetric bond set, and asserts the fixture is asymmetric so the check cannot go vacuous. It also covers the other consumers:

- **FLEX** — the reduced/squashed path contracts the same `ham_inter_q` down to the density block, so it inherits the orientation; every existing exact FLEX index-order test uses on-site interactions, where the transpose is invisible.
- **`trans_mod`** — `ham_inter_r` feeds `_calc_trans_mod` on the `green_init` path. That contraction sums over all `r`, i.e. it only sees `q = 0`, and for a bond set declared from both ends `Σ_r V_ab(r) = Σ_r V_ba(r)`, so this change must leave `trans_mod` alone. Asserted by running the same contraction on the interaction and on its orbital-pair transpose, with a guard that the two tensors really do differ.
- **Every interaction type** — `CoulombIntra`, `CoulombInter`, `Hund`, `Ising`, `PairLift` and `Exchange` share one placement statement, so they must share the orientation. `PairHop` is excluded on purpose: it is wired separately with a different four-operator mapping and restricted to `irvec == (0,0,0)`.

931 pass. The orientation assertions were verified to fail against the unfixed `rpa.py`; the `trans_mod` invariance passes on both sides, which is the point of it.

## Compatibility

Only orbital-asymmetric interactions change. Symmetric ones — including every fixture in the suite other than the 2-orbital RPA one — are bit-identical, and `trans_mod` is unaffected in all cases.

## Also in this PR

A note added in #94 to `tests/flex_bruteforce_ref.py` implied that MYO Eq.(5)'s verbatim form was simply wrong. It is a mapping problem: the paper's Green function index order is not this codebase's, so transcribing its equations term by term does not carry the convention with them. What is pinned there is the *composition* — the pair (bubble, Eq.(3)) must reproduce the exact O(U²) self-energy — not either equation alone. Reworded.

### DIFF
--- a/src/hwave/solver/rpa.py
+++ b/src/hwave/solver/rpa.py
@@ -466,7 +466,31 @@ class Interaction:
                 for spinvec, w in spins.items():
                     s1,s2,s3,s4 = spinvec
                     # beta beta' alpha alpha'
-                    orb = (s4, b, s3, b, s1, a, s2, a)
+                    # Orbital a -- the FIRST index of the input entry, i.e. the
+                    # one carrying momentum transfer +q -- goes in the beta
+                    # slot, so that after the e^{-iqR} FFT the reduced vertex is
+                    # V_ab(q) and not its orbital transpose (issue #96).
+                    #
+                    # Fixed by Mizuno, Kobayashi and Suzumura, arXiv:1010.1084,
+                    # which is the multi-site extended-Hubbard RPA this code
+                    # generalizes (the on-site-only sources cited elsewhere
+                    # cannot settle it: their vertex matrices are symmetric in
+                    # the orbital pair, so the orientation is invisible there):
+                    #   Eq.(4)  V_ab(q) a+_{k+q s a} a+_{k'-q s' b} a_{k' s' b}
+                    #           a_{k s a}          -> index a sits at +q
+                    #   Eq.(13) chi0_ab(q) = -(T/N) sum_k G_ab(k+q) G_ba(k)
+                    #                              -> the convention used here
+                    #   Eq.(12) chi^C = (I + 2 chi0 V + chi0 U)^-1 chi0, with V
+                    #           UNTRANSPOSED ("the hat denotes a matrix
+                    #           representation with an element (f)_ab = f_ab")
+                    # Their Eq.(6) also shows V_ab(q) is genuinely complex, with
+                    # V_ba = conj(V_ab), whenever the bond geometry is not
+                    # symmetric under a <-> b -- which is exactly when the
+                    # orientation matters.
+                    # Every spin table above is symmetric under exchanging the
+                    # (s1,s2) and (s3,s4) pairs, so moving the orbitals alone
+                    # transposes the vertex exactly.
+                    orb = (s4, a, s3, a, s1, b, s2, b)
                     ham_r[(*irvec, *orb)] += v * w
 
         # pairhop-type interaction
--- a/src/hwave/solver/rpa.py
+++ b/src/hwave/solver/rpa.py
@@ -466,31 +466,7 @@ class Interaction:
                 for spinvec, w in spins.items():
                     s1,s2,s3,s4 = spinvec
                     # beta beta' alpha alpha'
-                    # Orbital a -- the FIRST index of the input entry, i.e. the
-                    # one carrying momentum transfer +q -- goes in the beta
-                    # slot, so that after the e^{-iqR} FFT the reduced vertex is
-                    # V_ab(q) and not its orbital transpose (issue #96).
-                    #
-                    # Fixed by Mizuno, Kobayashi and Suzumura, arXiv:1010.1084,
-                    # which is the multi-site extended-Hubbard RPA this code
-                    # generalizes (the on-site-only sources cited elsewhere
-                    # cannot settle it: their vertex matrices are symmetric in
-                    # the orbital pair, so the orientation is invisible there):
-                    #   Eq.(4)  V_ab(q) a+_{k+q s a} a+_{k'-q s' b} a_{k' s' b}
-                    #           a_{k s a}          -> index a sits at +q
-                    #   Eq.(13) chi0_ab(q) = -(T/N) sum_k G_ab(k+q) G_ba(k)
-                    #                              -> the convention used here
-                    #   Eq.(12) chi^C = (I + 2 chi0 V + chi0 U)^-1 chi0, with V
-                    #           UNTRANSPOSED ("the hat denotes a matrix
-                    #           representation with an element (f)_ab = f_ab")
-                    # Their Eq.(6) also shows V_ab(q) is genuinely complex, with
-                    # V_ba = conj(V_ab), whenever the bond geometry is not
-                    # symmetric under a <-> b -- which is exactly when the
-                    # orientation matters.
-                    # Every spin table above is symmetric under exchanging the
-                    # (s1,s2) and (s3,s4) pairs, so moving the orbitals alone
-                    # transposes the vertex exactly.
-                    orb = (s4, a, s3, a, s1, b, s2, b)
+                    orb = (s4, b, s3, b, s1, a, s2, a)
                     ham_r[(*irvec, *orb)] += v * w
 
         # pairhop-type interaction
--- a/tests/flex_bruteforce_ref.py
+++ b/tests/flex_bruteforce_ref.py
@@ -29,15 +29,23 @@ The index slots below match these formulas exactly; do not "optimize" the
 wiring.
 
 Note on the bubble's index slots (issue #91).  These two formulas are a
-*pair*: only one relative orbital-pair order of the bubble reproduces the
-exact second-order self-energy when the two are composed as V = U chi0 U.
-This file previously carried the bubble with both pairs transposed
-(``G_{mu m} G_{n nu}``).  Composed with Eq.(3) that gives, for a density-only
-U, ``Sigma_{mn} ~ chi0_{nm} G_{mn}`` instead of ``chi0_{mn} G_{mn}`` -- a
-transpose that is invisible on the orbital diagonal and wrong by ~40% off it.
-The order below is the one pinned against an exactly-diagonalized 3-orbital
-model in ``tests/test_flex_sopt_index_order.py``, which is independent of any
-convention choice made inside this codebase.
+*pair*: only one relative orbital-pair order of the bubble reproduces the exact
+second-order self-energy when the two are composed as V = U chi0 U.
+
+The bubble is deliberately NOT the verbatim MYO Eq.(5), which reads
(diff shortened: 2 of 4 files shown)


k-yoshimi:
Converting to draft: this branch does not contain the production change it describes.

`git diff develop..HEAD` on it touches only three test files. The `rpa.py` edit was reverted by my own mistake — during a pre-fix verification I ran `git checkout develop -- src/hwave/solver/rpa.py`, which stages the reverted file, and the following commit picked it up. CI is failing accordingly, since the added tests expect the fix.

Separately, an adversarial review has challenged the scientific justification itself (whether H-wave's `W[beta,beta',alpha,alpha']` storage means the mathematical `V_ab` belongs at row b rather than row a). That needs to be settled against an operator-level calculation before this is re-proposed, not by index reasoning — see #96.

k-yoshimi:
Closing: the change in this PR is wrong. The original `_make_ham_inter` placement is correct.

Settled by exact diagonalization, with no reference to any convention in this codebase. The construction puts the interaction between opposite spins only,

    H_int = V sum_{i,R,a,b} v_ab(R) n_{i a up} n_{i+R b down}

so there is no exchange diagram (different species) and the cross-spin response `chi^ud_ab(q) = <rho^u_a(q) rho^d_b(-q)>` vanishes identically at V = 0 — verified, 2.8e-16 at q != 0. Its leading term is therefore exactly one bubble-vertex-bubble chain, with nothing else mixed in.

Measured on a 3-site ring, 2 orbitals, with `v_ab(R) != v_ba(R)`:

| | residual vs `chi0 v chi0` | residual vs `chi0 v^T chi0` |
|---|---|---|
| dV = 1e-3 | 9.84e-01 | 2.84e-05 |
| dV = 1e-4 | 9.84e-01 | 2.84e-06 |
| dV = 1e-5 | 9.84e-01 | 2.84e-07 |

The residual against `v^T` scales linearly in dV, i.e. it is pure O(V^2) truncation; against `v` it is flat at 98%. So the matrix entering `chi = [I + chi0 @ W]^{-1} chi0` is `W = V(q)^T`, which is what `_make_ham_inter` has always built.

That also vindicates `RPATwoOrbital`'s `.T`: my reading of edcf5cd as an ad-hoc patch was wrong, and the review's counter-reading — that it was a coherent correction matching the solver's beta-first storage after the fixture bonds were mirrored — is the right one.

#96 is not closed by this. The two paths do build different orientations, and that part of the report stands; what changes is which side needs correcting. See #96.

PR #102, opened 2026-07-28
Settle PairHop's vertex orientation by end-to-end measurement

### LINKED ISSUE #100: PairHop's vertex orientation is undetermined (excluded from the #96 transpose)
Split out of the #96 review (PR #99), which deliberately did not change PairHop.

## The gap

`sc.py::_build_interaction_k` now stores the orbital pair transposed for every interaction type **except** `PairHop`, and that exception is a decision made on the absence of evidence rather than on evidence.

The determination behind #96 is for a **density-density** term: `V_ab(R) n_ia n_{i+R,b}`. In `rpa.py` all of `CoulombIntra`, `CoulombInter`, `Hund`, `Ising`, `PairLift` and `Exchange` are placed by `_append_inter` with the slots

```python
orb = (s4, b, s3, b, s1, a, s2, a)      # beta beta' alpha alpha'
```

so the result transfers to them directly. `PairHop` is the only one placed by `_append_pairhop`, with **different** slots:

```python
orb = (s4, b, s3, a, s1, a, s2, b)
```

It is also the only type restricted to `irvec == (0,0,0)`, and its documented operator is a directed pair transfer:

```
H^PH = P^{aa'} c+_{i a up} c_{j a' up} c+_{i a down} c_{j a' down} + (up <-> down)
```

So a wrong orientation here would reverse the direction of the pair transfer, not merely relabel a density.

## Why it has not bitten

`P_ab` is symmetric in the orbital indices for every fixture in the tree, and for any physically ordinary input. It matters only for an input where `P_ab != P_ba`, which the readers accept without validation (#93).

## What to do

Determine it the way #96 was determined, with an exact calculation rather than by extending the density-density argument:

- the exact two-particle matrix element for the pair transfer, in the two-electron sector, on an asymmetric (and ideally complex) `PairHop` input;
- compared against the `(ab),(ba)` block that `_build_sc_matrices_all_q` Case 4 fills from `PH_mat[l1,l2]`, and against `rpa.py`'s `_append_pairhop` placement.

`tests/test_vertex_orientation.py` already has the machinery; `TestPairingVertexOrientation` is the closest template. The current behaviour (not transposed) is pinned there by `test_pairhop_is_deliberately_left_alone`, so whichever way it goes the change is a one-line edit plus that test.

Related: #96, #93.

k-yoshimi:
Closes #100.

**Outcome: the existing placement is correct.** `_to_k` still leaves PairHop alone, and `sc.py` changes only in comments. No numerical output changes.

An earlier revision of this branch transposed PairHop. That was wrong, and how it went wrong is why this is now settled differently.

### The two measurements, which are sound and are kept

**`TestChi0PairIndexMeaning`.** `rpa.py::_calc_chi0q` builds, in the general branch, `chi0[a,c,b,d] = G[a,b] G_rev[d,c] sgn`, and the S/C matrices are indexed by the flattened pairs. Which *bilinear* does each pair stand for? Comparing the kernel against the exact correlator of the generalised densities `A_ab(q) = sum_i e^{-iqr} c+_ia c_ib`, with G from the same diagonalization, singles out one of the 24 index permutations:

```
chi0[a,c,b,d](q)  ~  <A_ca(q) A_bd(-q)>
```

The **row** pair is reversed relative to its bilinear; the column pair is not.

**`TestPairHopVertexPlacement`.** `H_PH = sum_{aa'} P_aa' A^u_aa' A^d_aa'` couples an up bilinear to a down one, so the cross-spin response vanishes identically at `P = 0` and its derivative is exactly one bubble-vertex-bubble chain. Inverting it gives a vertex **diagonal in the bilinear pair**, with `Gamma[(a,b),(a,b)] = P[a,b]` exactly — sign and normalisation included, not merely proportional.

### What was wrong: combining them by hand

The vertex sits *between* two chi0 factors, so its row index lives in chi0's **column** convention and its column index in chi0's **row** convention. The measured reversal is on the row pair, so it lands on the vertex's **column** side. Applying it to the row side instead yields the transpose — which is what the earlier revision did.

Two derivations of the same shape reached opposite conclusions. That is reason enough not to settle it by derivation at all.

### What settles it: `TestPairHopEndToEnd`

No index reasoning anywhere between the input file and the answer. The chain under test is the one the solver runs:

```
interaction dict -> _build_interaction_k -> _build_sc_matrices_all_q -> S
chi0             <- rpa.py's own kernel, fed an EXACT Green function
prediction:  chi^ud = -chi0 . S . chi0        (linear order)
```

compared against the cross-spin response of the same Hamiltonian by exact diagonalization. Both orientations are run, and the test states which placement each one is, so it cannot silently invert if `_to_k` changes.

| placement | residual at 192 tau points | at 384 | at 768 |
|---|---|---|---|
| `S[(a,b),(b,a)] = P[a,b]` (current) | 2.3e-3 | 1.1e-3 | 5.5e-4 |
| `S[(a,b),(b,a)] = P[b,a]` (transposed) | 0.551 | 0.551 | 0.551 |

The winner tracks chi0's own imaginary-time discretisation — the test measures that separately in the same run, by checking chi0 against the exact bubble, so a larger residual downstream would be a real disagreement rather than the grid. (The test itself runs the first column; the refinement was measured off-line.)

### Also addressed

- The PairHop Hamiltonian no longer symmetrises `up @ dn`. The two bilinears commute, so symmetrising put a spurious factor of two into the measured coefficient; the vertex coefficient is now asserted exactly.
- The central-difference truncation is documented as O(dP^2), not O(dP^3), with a dP-halving test and a condition-number bound on X0 instead of an asserted tolerance.
- The fixture is complex and Hermitian-closed (`P_a'a = conj(P_aa')`). A real asymmetric PairHop matrix makes the Hamiltonian non-Hermitian, so it cannot be used for this at all; a complex Hermitian one still has `P^T = conj(P) != P`, which is what makes it discriminating.

942 tests pass.

### DIFF
--- a/src/hwave/sc.py
+++ b/src/hwave/sc.py
@@ -589,7 +589,7 @@ def _build_interaction_k(kx_array, ky_array, kz_array, interactions, norb):
         kx_array, ky_array, kz_array, indexing='ij'
     )
 
-    def _to_k(value_r, transpose=True):
+    def _to_k(value_r):
         # Same Fourier phase as the solver core, e^{-iqR}, but the ORBITAL
         # PAIR is stored transposed: an entry (R, (a, b)) lands at [b, a].
         #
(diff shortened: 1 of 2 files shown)


k-yoshimi:
Converting to draft: the composition step is not safe.

The two measurements stand -- chi0's row pair is reversed relative to its bilinear, and the PairHop vertex is diagonal in the bilinear pair. What does not stand is the step that combines them.

The vertex sits BETWEEN two chi0 factors in `[I + chi0 U]^{-1} chi0`, so `U`'s row index lives in chi0's COLUMN convention and `U`'s column index lives in chi0's ROW convention. The reversal was measured on the row pair, so it applies to `U`'s COLUMN side, not its row side. Carrying that through gives `U[(l1,l2),(l2,l1)] = P[l1,l2]` -- the opposite of what this branch implements.

Two derivations of the same shape have now reached opposite conclusions, which is reason enough not to settle it by derivation at all. Replacing the composition with a single end-to-end measurement: predict the cross-spin response from the code's own chi0 and the builder's S/C matrices, and compare it against exact diagonalization of the same Hamiltonian, for both orientations. That leaves no index reasoning between the input file and the answer.

Issue #90, opened 2026-07-27
RPA transverse (ladder) channel drops the CoulombInter contribution to the vertex


k-yoshimi:
Found while trying to add a cross-check between the transverse and longitudinal spin channels (#89, since closed). The cross-check itself was wrong, but chasing why exposed a defect in the shipped transverse channel.

## Symptom

`RPA._build_transverse_channel` builds `ham_pm` with no `CoulombInter` contribution — its own comment table says `CoulombInter V -> W_+- = 0`. The longitudinal builder gives `U'` a nonzero case-2 entry, and the two cannot both be right.

Measured on a two-orbital on-site model, `U = 2.0`, `U' = 0.7`, `calc_scheme="general"`, spin-free:

```
max|S|            = 2.0000        longitudinal spin vertex
max|ham_pm|       = 2.0000        transverse vertex as built
max|ham_pm + S^T| = 0.7000        should be 0

at q=0, off-diagonal orbital-pair index (0,1):
  S[1,1]      = +0.7000     longitudinal, = U'
  ham_pm[1,1] =  0.0000     transverse, as the code builds it
  -S^T[1,1]   = -0.7000     transverse, as derived below
```

The discrepancy is exactly `U'`. `CoulombIntra`-only models are unaffected, because that vertex is diagonal in the orbital-pair index and invariant under the transpose at issue — which is why existing tests do not catch this.

## Derivation

The public bare bubble is the pair matrix `X_R[(a,c),(b,d)] = chi0[a,c,b,d]`. `FLEX._flex_compute_veff_general` converts it to MYO order as `X_M = X_R^T` (`transpose(0,1,4,5,2,3)`), solves `chi_s^M = [I - X_M S]^{-1} X_M`, and transposes back. Transposing that matrix equation gives

```
chi_s^public = [I - X_R S^T]^{-1} X_R
```

`_solve_rpa` computes `[I + X_R W]^{-1} X_R`, so the transverse vertex required in public order is

```
W_+- = -S^T          i.e.   W_+-[a,c,b,d] = -S[(b,d),(a,c)]
```

With that vertex the elementwise identity is `chi_pm == chiq_s` with factor 1. (The familiar `chi^{+-} = 2 chi^{zz}` uses `S^z = (n_up - n_down)/2`; the stored `chiq_s` omits that 1/2, so it already equals the code's `chi^{+-}`.)

[...]

PR #103, opened 2026-07-28
Abandoned: transverse (ladder) vertex — the proposed fix was the wrong direction

k-yoshimi:
**Closed without merging. The production change here is wrong and must not be applied.** The findings are recorded in #90, which stays open.

## What this branch proposed

`_build_transverse_channel` builds the ladder vertex as `up_block - cross_block.transpose(0,4,3,2,1)`. This branch dropped the transpose, making `CoulombInter` cancel exactly between the two spin blocks — matching the comment table in the method, which says `CoulombInter V -> W_+- = 0`.

That restored `chi_zz = chi_pm` for a two-orbital CoulombInter, from 2.9e-2 to 1.0e-16, and I took the agreement as confirmation.

## Why it is wrong

Exact diagonalization, with the O(lambda) self-energy subtracted exactly via the response to the Hartree-Fock one-body potential, gives a q-independent transverse vertex of **-V** for on-site CoulombInter, not zero. Numbers and method in #90.

The oracle was the problem. `_assert_su2` compares the **ring** `chi_zz` against the **ring+ladder** `chi_pm` and demands equality. Those sit at different diagrammatic levels — the ring puts a spin-independent density interaction entirely in the charge channel and leaves spin bare (measured: spin vertex 7.4e-15, charge vertex 4.0), while the exchange contribution that also feeds the spin channel is exactly what a ladder is for. So the equality is only satisfiable by making the ladder contribute nothing for V, and this branch achieved it by doing precisely that. The table entry `CoulombInter -> W_+- = 0` appears to have been derived from that test rather than from the physics.

## What still stands, and is not addressed here

1. **The shipped vertex is wrong under any reading.** It evaluates to `V(q) - V(q)^T`, purely antisymmetric — matching neither 0 nor -V. It cancels only when V(q) is symmetric under the pair transpose.
2. **The test suite cannot see pair-index errors.** Every CoulombInter SU(2) case runs on a one-orbital fixture, where the orbital-pair index takes a single value; and `_assert_su2` compared only the pair-diagonal elements at norb > 1, which are invariant under the transpose at issue.
3. **The SU(2) oracle is not valid as written**, per above.
4. **PairHop is the only interaction type that separates the tied index permutations**, and it was absent from the search here.

The tests added on this branch encode the wrong conclusion and are discarded with it. Items 1-4 belong to #90.

### DIFF
--- a/src/hwave/solver/rpa.py
+++ b/src/hwave/solver/rpa.py
@@ -2218,6 +2218,8 @@ class RPA:
         # Verification for each interaction type:
         #   CoulombIntra U: ham[↑↑↑↑]=0, ham[↓↓↑↑]=U  → W_+-= -U  (correct)
         #   CoulombInter V: ham[↑↑↑↑]=V, ham[↓↓↑↑]=V  → W_+-=  0  (SU(2) correct)
+        #                   -- and the cancellation is only exact if the two
+        #                      blocks are taken in the SAME index order; see #90
         #   Exchange J:     ham[↑↑↑↑]=0, ham[↓↓↑↑]=0   → W_+-=  0  (SU(2) correct)
         #   PairLift J:     ham[↑↑↑↑]=0, ham[↓↓↑↑]=0   → W_+-=  0  (SU(2) correct)
         #   Hund J:         ham[↑↑↑↑]=-J, ham[↓↓↑↑]=0  → W_+-= -J  (not SU(2) alone)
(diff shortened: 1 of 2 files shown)


k-yoshimi:
Added a check for a premise the vertex extraction was resting on, raised in review.

The vertex the transverse channel needs was obtained by inverting the equation the solver runs,

```
chi = [I + chi0 W]^-1 chi0    =>    W = chi^-1 - chi0^-1
```

and then substituting the LONGITUDINAL `chi_zz` for `chi` while keeping the TRANSVERSE `chi0_pm` for `chi0`. Those are two different channels. As a *definition* — "what W would make the transverse solve output chi_zz" — the substitution is exact and needs nothing. Reading the result as a vertex assumes something extra: that `chi_zz` itself satisfies a `norb^2 x norb^2` equation with bubble `chi0_pm`. The longitudinal solve is a `(norb*2)^2` problem in spin-orbital space and `chi_zz` is a projection of it onto the spin sector; nothing shown so far established that the projection closes.

Now measured rather than assumed. For an on-site interaction the vertex is q-independent, so the extracted W must be too — if the projection did not close, no single q-independent W could reproduce the projected object.

| interaction | `|W|` | spread over q |
|---|---|---|
| CoulombIntra (on-site U = 4) | 4.0000 | 2.7e-15 |
| Hund (on-site) | 1.0000 | 5.2e-15 |

q-independent to machine precision, with `|W|` equal to the interaction strength itself. The test also asserts the vertex is nonzero, so it cannot pass vacuously on a case where W = 0 makes q-independence trivial.

Worth noting where this does and does not matter. It matters for the cases with a **nonzero** vertex, i.e. the permutation search. The CoulombInter conclusion does not rest on it at all: `W_needed ~ 0` there is the direct statement that `chi_zz` equals `chi0_pm`, a comparison of two measured arrays with no RPA-form assumption. SU(2) then forces `chi_pm = chi_zz = chi0_pm`, and requiring `[I + chi0_pm W]^-1 chi0_pm = chi0_pm` gives `W = 0` outright (`chi0_pm` invertible, cond 1.02).

Also corrected the module docstring, which stated the chi_zz = chi_pm identity without the same-index-order proviso that #90 turns out to hinge on.

945 tests pass.

k-yoshimi:
Converting to draft. A review raised three blocking points, two of which I had just confirmed independently by measurement.

### 1. The residual permutation freedom IS reachable — via PairHop

The source comment claims the four tied candidates cannot be separated by any reachable input. That is wrong. Measured, on a two-orbital fixture with an asymmetric on-site entry, comparing each tied permutation against the plain block:

| interaction | perm (1,2,4,3) | (2,1,3,4) | (2,1,4,3) |
|---|---|---|---|
| CoulombInter | 0 | 0 | 0 |
| Hund | 0 | 0 | 0 |
| Exchange | 0 | 0 | 0 |
| PairLift | 0 | 0 | 0 |
| Ising | 0 | 0 | 0 |
| **PairHop** | **1.0** | **1.0** | **0.65** |

PairHop is placed by `_append_pairhop` with different slots and is the one type that separates them — and it was absent from the search. The claim has to go, and PairHop's slot is undetermined.

### 2. The existing SU(2) oracle is not a valid SU(2) oracle

The suite asserts `chi_zz == chi_pm` for Ising and for Hund alone, while the vertex table in `rpa.py` says both are "not SU(2) alone". So those assertions enforce the implementation's assumed block identity rather than a symmetry, and a systematically wrong vertex can pass them.

Redoing the permutation search restricted to genuinely SU(2)-symmetric input (CoulombIntra, CoulombInter, Intra+Inter, full Kanamori) leaves the conclusion unchanged — the same four tie at 7.4e-3, everything else including the previous code sits at 2.0e12 — so the fix direction does not depend on the invalid cases. But the tests do need correcting.

### 3. The open one: is matching the longitudinal channel the right criterion at all?

The review argues that `up_block - cross_block` erases CoulombInter identically, whereas a transverse ladder should still carry a Fock kernel on orbital-changing spin-flip pairs — i.e. that issue #90's original diagnosis applies in a regime the longitudinal projection cannot see, and that selecting the vertex by matching the longitudinal ring solve is circular.

The counter-argument is that chi^{+-} = chi^{zz} is an exact identity for an SU(2)-symmetric Hamiltonian, so any approximation respecting SU(2) must satisfy it, whatever else it keeps. That would make the criterion valid regardless of completeness.

I am not settling this by argument. Both readings are the same shape of reasoning that has reversed twice in this series. What is needed is an independent determination of the transverse vertex that does not use the longitudinal solve as its reference.

### What is not in doubt

The defect itself. For two-orbital CoulombInter — an interaction that IS SU(2)-symmetric — the shipped code breaks chi_zz = chi_pm by 2.9e-2 on a scale of 1.2e-1, and the vertex it builds is purely antisymmetric with its symmetric part exactly zero. Whatever the correct vertex turns out to be, the current one is wrong.

k-yoshimi:
The exact-diagonalization determination is done, and it reverses this branch. Full numbers in #90.

In short: an ED extraction that subtracts the O(lambda) self-energy exactly (via the response to the Hartree-Fock one-body potential) gives a q-independent transverse vertex of **-V** for on-site CoulombInter, not zero. The independent check is that the same-spin part of V alone gives exactly zero vertex -- it cannot connect the up and down propagators of the transverse loop, so it can only produce self-energy -- while the cross-spin part gives -V. That check also caught a transposed index in the Fock term of my own extraction script, which had produced a spurious 5e-2 of q-dependence and led me to report, wrongly, that the exchange contribution was not representable as a static vertex. It is.

So the production change here should not be merged. Setting the vertex to zero does make chi_pm agree with chi_zz, but only because the ring-level chi_zz omits the same exchange physics; the ladder then contributes nothing for V, which is what the channel exists to add.

What survives from this branch, and is worth keeping independently of the vertex decision:

- the shipped vertex evaluates to V(q) - V(q)^T, purely antisymmetric, which matches neither candidate -- it is wrong under any reading;
- `_assert_su2` compared only the pair-diagonal elements at norb > 1, and every CoulombInter case ran on a one-orbital fixture, so nothing in the suite could see a pair-index error;
- the SU(2) oracle itself is not valid as written: it compares a ring chi_zz against a ring+ladder chi_pm and demands equality, which is only satisfiable by a trivial ladder.

I will not push a vertex change without a decision on the third point in #90, since it changes what chiq_pm means.

Issue #104, opened 2026-07-28
RPA ring drops the inter-orbital term from the spin vertex, disagreeing with the FLEX path and with exact diagonalization


k-yoshimi:
## Summary

`rpa.py`'s ring solve builds its spin susceptibility from a vertex that has no inter-orbital term. For `CoulombIntra` U plus on-site `CoulombInter` U' on two orbitals, the effective spin vertex extracted from the solver's own output is

```
rpa.py ring :  diag(U_0,  0 ,  0 , U_1)
sc.py S     :  diag(U_0, U', U', U_1)
```

The `(a,b),(a,b)` slots -- the orbital-off-diagonal spin fluctuations -- are simply absent. `sc.py::_build_sc_matrices_all_q` fills them (Case 3), and exact diagonalization agrees with `sc.py`, so the two paths in this repository disagree about the same physical vertex and the RPA one is the deficient side.

This is not a different resummation of the same physics. The term is already present at **first order**, and the ring drops it.

## How it was measured

Two independent routes, neither of which uses the other as its reference.

**1. What the solver actually uses.** From `chi = [I + chi0 W]^-1 chi0`, which is what `_solve_rpa` computes, the vertex follows by inversion: `W = chi^-1 - chi0^-1`. Applied to the spin projection `chi_uu - chi_ud` of the solver's own `chiq`, on a two-orbital fixture with U = (2.0, 1.3) and on-site U' = 0.7:

```
effective spin vertex = diag(2.000, 0, 0, 1.300)      q-spread 3.6e-15
```

The vertex is q-independent to machine precision, as it must be for an on-site interaction -- so the extraction is clean, and the projection onto the spin sector closes.

**2. What it should be.** Exact diagonalization of a 3-site, 2-orbital, spinful model. The obstacle to comparing ED with RPA is that RPA is defined by which diagrams it keeps, so the exact O(lambda) response also carries the O(lambda) self-energy dressing of the propagators, which the solver does not have (it builds chi0 from the non-interacting Green function). At O(lambda) that dressing is exactly the linear response to the Hartree-Fock one-body potential built from the lambda = 0 density matrix, so subtracting it leaves the vertex alone.

Comparing the two candidate vertices against that reference, through the solver's own `-chi0 . W . chi0`:

| vertex | 192 tau points | 384 | 768 |
|---|---|---|---|
| `diag(U_0, 0, 0, U_1)` — what the ring uses | 0.560 | 0.561 | 0.561 |
[...]

Issue #93, opened 2026-07-28
Interaction files are never checked for orbital symmetry, and each solver path handles an asymmetric file differently


k-yoshimi:
**Priority: medium.** Not a wrong-result bug on valid input, but every solver silently accepts input it cannot mean, and each one does something different with it. Surfaced while fixing #91, where an asymmetric interaction file turned out to be the one case in which that fix changes the saved susceptibilities.

## The invariant that is never checked

The two-body terms are density–density (or pair) couplings between orbitals:

```
V_ab n_a n_b ,  J_ab S_a . S_b ,  ...
```

so `X[a,b]` and `X[b,a]` name the *same* coupling. An input file that declares them with different values is not a different physical model — it is a typo. Nothing rejects it.

Concretely, `wan90.read_w90` stores `(irvec, orbvec) -> value` verbatim; its only checks are the degeneracy count, duplicate keys, and dense-listing (`src/hwave/qlmsio/wan90.py:202-257`). `read_input.py` has the same shape, and carries a standing `# TODO Check Pair(Hermite or not)` at `src/hwave/qlmsio/read_input.py:23`.

## Every consumer handles it differently

| path | behaviour on `X[a,b] != X[b,a]` |
|---|---|
| `uhfk.py` `_make_ham_inter` | **silently projects** onto `(J_ab(r) + conj(J_ba(-r)))/2` — `uhfk.py:1109,1153,1179,1205,1233,1259,1285`. No warning. `CoulombIntra` is not projected (`uhfk.py:1090,1132`); it is instead index-constrained by `_check_orbital_index` (`uhfk.py:731-733`), and in the explicit branch off-site / `alpha != beta` entries are silently dropped (`uhfk.py:1128-1129`) |
| `rpa.py` `Interaction._make_ham_inter` | **verbatim** — `_append_inter` writes only the declared ordering (`rpa.py:459-470`), `_append_pairhop` likewise (`rpa.py:477-487`). Asymmetric `ham_inter_q` |
| `sc.py` `_build_interaction_k` / `_build_sc_matrices_all_q` | **verbatim** — `_to_k` at `sc.py:1247-1257`; the S/C builder reads only `X[l1,l2]` / `X[l1,l3]` (`sc.py:1449-1491`) |
| `_sc_matrices_myo.py` `build_sc_matrices_myo` | **verbatim** — same structure (`_sc_matrices_myo.py:79-123`); consumed by the FLEX `general` path (`flex.py:1989-1993`) |
[...]

PR #105, opened 2026-07-28
Rebuild the RPA transverse (ladder) vertex against exact diagonalization

### LINKED ISSUE #90: RPA transverse (ladder) channel drops the CoulombInter contribution to the vertex

k-yoshimi:
Closes #90.

`_build_transverse_channel` built the vertex as `up_block - cross_block.transpose(0, 4, 3, 2, 1)`, and the table of expected values above it had **four of seven entries wrong and one missing**. Every value here comes from exact diagonalization of an explicit 3-site, 2-orbital, spinful model, with the O(lambda) self-energy removed exactly — it is the linear response to the Hartree-Fock one-body potential built from the non-interacting density matrix, so subtracting it leaves the vertex alone.

An earlier attempt at this issue (#103) proposed the opposite fix and was wrong; that history is in #90.

## What was wrong

**The same-spin block must not appear.** A same-spin interaction cannot connect the up and down propagators of the transverse loop, so it produces only self-energy and no vertex — measured 1.4e-5 against 0.63 for the cross-spin part. `up_block` is exactly that block.

**Exchange was dropped entirely.** Measured block occupancy:

```
CoulombIntra   uudd, dduu          Hund       uuuu, dddd
CoulombInter   uuuu/uudd/dduu/dddd Exchange   udud, dudu
Ising          uuuu/uudd/dduu/dddd PairLift   uddu, duud
PairHop        uudd, dduu
```

Exchange and PairLift sit outside the cross-spin block. PairLift is harmless — its transverse vertex vanishes identically — but Exchange is not.

**Exchange must be symmetrised in the orbital pair.** `X_ab† = X_ba`, so `H = Σ J_ab (X_ab + X_ba)` depends only on `J_ab + J_ba`. For an asymmetric real `J = (1.0, 0.35)` — physical, since Exchange is Hermitian for any real J — the required vertex is `-(1.0 + 0.35)` at both slots, while the block holds `-0.35` and `-1.00` separately.

## The result

```python
ham_pm = (-cross_block.transpose(0, 2, 4, 1, 3)
          + spin_flip_block + spin_flip_block.transpose(0, 3, 4, 1, 2))
```

| interaction | measured | previous table | |
|---|---|---|---|
| CoulombIntra | `-U` | `-U` | correct |
| CoulombInter | `-U'` | `0` | **wrong** |
| Hund | `0` | `-J` | **wrong** |
| Ising | `+J` | `2J` | **wrong** |
| Exchange | `-(J + J^T)` | `0` | **wrong** |
| PairLift | `0` | `0` | correct |
| PairHop | `-J` | *no entry* | **missing** |

Verified end to end for all seven types, in chi0's own labels, against the solver's own `-chi0 . W . chi0`. The ED-to-code scale comes out as **one constant across every type** (3.048 to 3.055, i.e. `N_site`), so signs and magnitudes are under test and not merely structure; every residual sits at the imaginary-time discretisation floor (1.1e-3 to 4.1e-3).

The orbital ordering is one of two the measurements cannot separate. They differ only for off-site interactions (now rejected) and for complex density-density coefficients, which make the Hamiltonian non-Hermitian and are not physical input — the readers accept them anyway, see #93. On every physically valid on-site input, including a complex Hermitian-closed PairHop, the two agree exactly.

## Off-site terms are now rejected, per type

The transverse pair is `c+_(i a up) c_(j b down)`; when the cross-spin part of the interaction is off-site its two ends sit on different sites, the pair is intrinsically non-local, and `ham_pm` — which carries only on-site orbital pairs — has nowhere to put it. Measured off-site, asking whether the extracted vertex is proportional to `V(q)` with one common constant:

| | off-site transverse vertex | |
|---|---|---|
| CoulombInter | 0.63, residual 1.0 | rejected |
| Ising | 0.63, residual 1.0 | rejected |
| Exchange | 2.4, residual 0.33 | rejected |
| Hund | 1.4e-5, i.e. zero | accepted |
| PairLift | exactly zero | accepted |
| PairHop | discarded off-site by `_append_pairhop` | accepted |

The guard is on the cross-spin and spin-flip blocks, not the whole tensor; a blanket guard would refuse Hund and PairLift needlessly. **The longitudinal channel is unaffected** — its vertex *is* `V(q)`, to 1e-8, which is ordinary extended-Hubbard RPA.

## Tests

- `_assert_su2` compared only the pair-**diagonal** elements at `norb > 1`. Those are invariant under the orbital-pair transpose, so it reported agreement while `chi_zz` and `chi_pm` differed by 9 percent. It now covers the full matrix.
- Seven SU(2) cases ran on the one-orbital fixture whose `coulombinter.dat` carries off-site bonds. They were never valid tests of this channel; replaced by guard tests plus on-site two-orbital cases.
- `test_su2_hund_2orb` asserted SU(2) for Hund alone, which is not SU(2) symmetric. It passed only because both channels were built from the same expression.
- New: the full vertex table pinned against exact diagonalization, and the Exchange symmetrisation pinned on an asymmetric fixture.

`test_su2_onsite_coulombinter_2orb` records that SU(2) is now **violated** for on-site CoulombInter, by 1.0e-2 on a scale of 1.2e-1. That is correct, and it detects **#104**: the longitudinal ring leaves the spin channel completely undressed (`chi_zz - chi0` is 2.8e-17) where `sc.py` and exact diagonalization both give `U'`. It must not be made to pass by removing `U'` from the transverse side as well.

## Effect on results

`chiq_pm` is an output only — it feeds neither `chiq`, the self-energy, nor the Eliashberg vertex. Only `calc_type="ring+ladder"` results change. No committed reference data changes.

940 tests pass; the new ones fail on the unfixed code.


k-yoshimi:
Converting to draft. A review found a real defect in this branch, and chasing it exposed a convention question that reaches beyond the transverse channel.

## The defect, confirmed

The cross-spin block is not symmetrised in the orbital pair, while the spin-flip block is. That is an inconsistency in this branch, not in the original code.

The interaction files carry a redundant representation: for a density-density term `n_a n_b = n_b n_a`, so only `V_ab + V_ba` is physical. The sharpest consequence, measured:

| input | Hamiltonian | `ham_pm` on this branch |
|---|---|---|
| on-site CoulombInter, `V_01 = +1`, `V_10 = -1` | identically **zero** | `(0,1),(0,1) = +1.00`, `(1,0),(1,0) = -1.00` |
| the same as Ising | identically **zero** | `-1.00`, `+1.00` |

So the response depends on how the user split the coefficient between two entries that denote the same operator.

The symmetrising transposes are identified — `(2,3,0,1)`, `(2,3,1,0)`, `(3,2,0,1)`, `(3,2,1,0)` all cancel the antisymmetric part and double the symmetric one.

## Why it is not simply applied

Symmetrising changes the magnitude for ordinary symmetric input, from `0.70` to `1.40`, so it forces a decision on what two file entries mean — the sum or the average. The evidence points both ways:

- `uhfk.py` takes the **average**: its Ising branch builds `(jab_r + jba)/2`.
- Exchange requires the **sum**: `X_ab^dagger = X_ba`, so `H = sum_ab J_ab (X_ab + X_ba) = (J_01 + J_10)(X_01 + X_10)`, and exact diagonalization on an asymmetric `J = (1.0, 0.35)` gives a vertex of `-1.35`, not `-0.675`.

Both cannot be right for the same file format, and picking one by guesswork is exactly what should not happen here.

## Same root as the review's second point

The review also flags that `rpa.py`'s Ising construction lacks the `1/4` that `uhfk.py` applies. That is the same disagreement between the two solvers about how to read an interaction file, seen from another side. It is pre-existing and independent of this branch, and fixing it changes `chiq`, not just `chiq_pm`.

(One part of that finding does not hold: the review states the documented Ising Hamiltonian is `J Sz_i Sz_j`. The file specification writes it as `sum J^z (n_up - n_dn)(n_up - n_dn)`, without the 1/4, which is what the diagonalization here was built to match.)

## Also outstanding

- `docs/en/source/algorithm/rpa.rst` still publishes the old vertex table (`CoulombInter -> 0`, `Hund -> -J`, `Ising -> 2I`) and states that all interaction types are supported. It is now the opposite of the behaviour and must be updated in every maintained language.
- The off-site guard raises after the longitudinal solve has already run and populated `green_info['chiq']`. `_make_ham_inter()` is called from `__init__`, so it can validate before any expensive work.
- Off-site PairHop is accepted only because `_append_pairhop` silently discards it, so a user can supply an unsupported term and get a calculation with it quietly missing.
- The spinful branch already drops spin-mixing cross terms with a warning; the new vertex is untested there.

What is settled and unaffected by the above: the same-spin block must not appear at all, Exchange enters through the spin-flip block, and the per-type off-site classification. Those rest on measurements that the symmetrisation question does not touch.

k-yoshimi:
Second review round complete; all findings addressed in `35abc07`.

**must_fix — confirmed independently before the review returned, then fixed.** The pair-swap average used a plain transpose; `uhfk.py` averages with the **conjugate** (`jba = conjugate(transpose(...))`), and the two coincide only for real input. Measured on a complex Hermitian-closed PairHop `P = 0.7 ± 0.4i`: the plain mean collapsed both slots to `Re(P)` (`ham_pm = -0.700 + 0.000i`), where the exact vertex (#100) carries the full complex value. With the conjugate the mean is an identity on Hermitian-closed input: `-0.700 ∓ 0.400i` at the two slots, pinned by a regression test whose load-bearing assertion is a nonzero imaginary part. Non-Hermitian declarations (real antisymmetric PairHop) are projected onto their Hermitian part rather than rejected — input validation is #93, recorded in the source.

**should_fix, all applied:**
- The off-site guard is now `_check_transverse_representable`, called **before** the longitudinal solve, so invalid input fails without burning the ring solve or leaving `chiq` populated.
- `test_su2_onsite_coulombinter_2orb` pins the specific #104 signature — mismatch in the off-orbital pair slots within the measured magnitude, density-diagonal block still agreeing to 1e-10 — instead of a bare `assertRaises`.
- Docs: the on-site restriction moved from the spin-orbital section to the transverse section where it belongs (EN/JA); the silent discard of off-site PairHop by `_append_pairhop` is now stated in a warning instead of being hidden by the 'every two-body term' phrasing; the Ising `+I` entry is labelled as RPA's current normalization with a pointer to the tracked factor-4 discrepancy against UHFk (#106).
- The antisymmetric-declaration-of-zero test now covers Exchange as well.

Deferred, tracked: spinful spin-mixing cross terms are still dropped with a warning (pre-existing, unchanged by this branch); the Ising normalization itself is #106; asymmetric-input validation is #93.

942 tests pass; the new complex-PairHop test fails on the previous commit.

k-yoshimi:
Third review round complete; all findings addressed in `70bb874`, plus one defect the review could not have seen (it reviewed the previous commit).

**Found by measurement, beyond the review's scope: the two blocks need DIFFERENT symmetrisation partners.** Round 2 conjugated the pair-swap average on both blocks. That is right for the cross-spin block — its occupied slots pair as `((a,b),(b,a))`, where the pair swap coincides with reversal within each pair, so the redundancy is the Hermitian closure and complex PairHop survives. It is wrong for the spin-flip block — its slots pair as `((a,a),(b,b))`, palindromic within each pair, so the redundancy is the plain identity `X_ab = X_ba` and the coupling `(J_01 + J_10)` is real for any Hermitian declaration. Measured: complex Hermitian-closed Exchange `J = 0.7 ± 0.4i` produced `ham_pm = -0.7 ∓ 0.4i` — a spurious imaginary part where the coupling is `-0.7`. The spin-flip block now averages without conjugation. Two complementary regression tests pin this: the Exchange one asserts the imaginary part **vanishes**, the PairHop one asserts it **survives**.

(The review's suggestion to assert conjugate complex values for Exchange would have canonised the artifact — the operator identity forces a real coupling there.)

**must_fix — the guard now validates the assembled vertex.** Vertex assembly is extracted into `_assemble_transverse_vertex` (one implementation, provenance consolidated, revision-stale comments removed), and `_check_transverse_representable` builds the same `ham_pm` the channel will use and checks its q-spread — so a declaration whose off-site parts cancel in the symmetrised sum is accepted, and cancellation inside a raw aggregate can no longer hide an individual term. The duplicate call inside `_build_transverse_channel` is gone; validation runs once, before the longitudinal solve.

**should_fix, applied:** the #104 detector masks the expected U' slots and asserts every other element agrees to 1e-10; the module docstring reflects the current coverage instead of the removed formula; the docs (EN/JA) state the per-block partners explicitly with the complex consequences of each.

943 tests pass.

k-yoshimi:
One more defect, found by interrogating the recurrence pattern itself rather than by a review round (`7fedfa9`).

A complex Hermitian-closed **CoulombInter or Ising** declaration — `V_10 = conj(V_01)`, a Hermitian Hamiltonian — got a spurious imaginary vertex (`-0.7 ∓ 0.4i` measured, physical coupling `-0.7`), for exactly the reason Exchange did one commit earlier: the two declarations multiply the same operator, so the coefficient is the real sum and the imaginary parts are inert. The per-**block** rule missed it because the density types share the cross-spin block with PairHop, just in different slots.

The correct granularity is the **slot family**, and it is now implemented as one mask-selected mean applied uniformly to both blocks:

| slot family | redundancy | mean | occupants |
|---|---|---|---|
| palindromic `((a,a),(b,b))` | same operator; coefficient = real sum | plain | density types, Exchange |
| heterogeneous `((a,b),(b,a))` | Hermitian-partner coefficients; phase physical | conjugated | PairHop |

All twelve sentinel cases verified: every real case unchanged; complex Hermitian CoulombInter / Ising / Exchange real; complex Hermitian PairHop keeps its phase; all three antisymmetric declarations zero. The three identical complex fixtures are consolidated into one, and the density-density case is pinned by a regression test whose load-bearing assertion is the vanishing imaginary part.

With this, the declaration-class space is exhaustively enumerated and pinned: {7 types} × {real symmetric, real asymmetric, real antisymmetric, complex Hermitian-closed} on-site, plus per-type off-site accept/reject. 944 tests pass.

k-yoshimi:
Fourth (final requested) review round; findings resolved in `d1c988a`.

**must_fix 1, confirmed and fixed — the off-site pair swap must reverse the displacement.** Swapping the two bilinears of an off-site term sends `r → -r`, so the same-operator partner of `V_ab(+r)` is `V_ba(-r)` — the swapped slot at `-q`, not at the same `q`. Measured before the fix, the guard's verdicts were exactly backwards:

| input | Hamiltonian | verdict before | after |
|---|---|---|---|
| `V_ab(+x)=+1, V_ba(+x)=-1` | **nonzero** | accepted, vertex 0 (physics silently dropped) | **rejected** |
| `V_ab(+x)=+1, V_ba(-x)=-1` | identically zero | rejected | **accepted, vertex 0** |

The palindromic-family mean now takes its partner from the `-q` batch (nvol index map, flattening order verified against `slot(-q) = conj(slot(q))` for real input). The heterogeneous family needs no reversal — conjugation at fixed `q` is the Fourier image of {conjugate, `r→-r`}, i.e. exactly the Hermitian redundancy. All nine on-site sentinels unchanged; both off-site verdicts pinned by a policy test.

**must_fix 2 — diagonal PairHop is a reader question, not a channel divergence.** Measured: the entry denotes `2P n_up n_dn` and `_append_pairhop` stores coefficient `P` at the density slot, so both channels see the same effective coupling and the vertex assembly introduces no inconsistency. The factor semantics of that degenerate entry now lives in #93, and the docs carry the caveat.

**should_fix:** the #104 detector keeps its lower bound (the detector) and loses the brittle upper bound; docs wording corrected from "per block" to "per interaction type / slot family". **optional:** a combined-declaration test pins that mixed types superpose linearly.

946 tests pass. With this, the declaration space is closed: {7 types} × {real sym / asym / antisym, complex Hermitian} on-site, per-type off-site accept/reject including displacement-aware redundancy, plus the degenerate diagonal case documented and tracked.

k-yoshimi:
Fifth review round; findings resolved in `a860e54`.

**must_fix — the guard's tolerance had an absolute floor.** `spread > 1e-10 * max(scale, 1.0)` accepted any purely q-dependent off-site vertex of amplitude below 1e-10 (100 percent q variation, silently used). Now purely relative — Fourier roundoff is ~1e-16 relative, so 1e-10 relative separates it at any amplitude — and pinned by a 1e-12-amplitude fixture that must raise.

**should_fix, both applied:**
- Sublattice folding is now under committed test, not just hand-verified: CellShape [4,4,1] and non-cubic [4,6,1] with SubShape [2,2,1], checking the on-site vertex (q-independent 0.7), the rejected nonzero off-site declaration, and the accepted identically-zero one.
- The Japanese docs were indeed missing the diagonal-PairHop caveat — an earlier edit had failed silently. Restored; EN and JA now agree.

**optional, adopted:** the correct SU(2) statement now exists as an `expectedFailure` test asserting `chi_zz = chi_pm` for on-site CoulombInter. It fails while #104 is open; a fix of #104 flips it to a loud unexpected-success. The slot-signature detector stays for localization.

Also recorded in #106 during this round: for the identically-zero off-site declaration, the transverse channel correctly assembles a zero vertex while the longitudinal ring — which reads the raw unsymmetrised tensor — dresses `chiq` (4.3e-3 spin, 2.6e-2 charge) with an interaction that is the zero operator. That inconsistency exists for every asymmetric declaration and is precisely #106's scope.

948 passed, 1 xfailed.

k-yoshimi:
Sixth review round; resolved in `f12f45a`.

**must_fix — a pre-existing defect in code this branch had not touched.** `_calc_chi0q_transverse` (the spin-diag transverse bubble, from the original channel implementation) built `G_down(-r,-tau)` by flipping axes AFTER flattening the lattice to nvol — and the flipped axes included the two orbital indices. Modular negation of the flat C-order index is not coordinate-wise negation on a multidimensional lattice, and orbital indices must not be reversed at all. Both errors are invisible for nd ≤ 2 (roll(-1)∘flip is the identity on an axis of length ≤ 2) and on 1D lattices — exactly what every existing fixture used. Measured at nd = 3 on a 3×2 lattice: disagreement of order unity (2.8). Fixed to mirror the longitudinal `_calc_chi0q`, and pinned against an independent trusted reference: with `G_up == G_dn` the transverse bubble must equal the longitudinal general chi0 (the ED-verified kernel of this series) — now exact (0.0), previously 2.8.

**should_fix, applied:** stale `chiq`/`chiq_pm` can no longer leak across solves on a reused green_info (cleared at computation start; pinned by a test that also documents a separate pre-existing external-chi0q defect it had to sidestep); the docs (EN/JA) now state the actual acceptance criterion — q-independence of the assembled vertex on the folded lattice, relative 1e-10, with its two deliberate consequences spelled out. **optional, applied:** spy subtests clear and assert their capture.

Not taken up, deliberately: the unreachable reduced-expansion branches inside `_build_transverse_channel` (defensive; removing them is a separate compatibility question) and a CUDA-runner exercise (no GPU in this environment; the ops used — fancy indexing, where, conj — are the same families already exercised by the existing GPU test on this channel).

950 passed, 1 xfailed.

Issue #107, opened 2026-07-29
Verify RPA == FLEX(IterationMax=1, no self-energy update) across the input matrix, then unify RPA onto the FLEX path


k-yoshimi:
## Proposal

RPA is the one-shot limit of FLEX: one iteration, no self-energy update, chi built from the bare Green function. The two are currently separate code paths — `FLEX(RPA)` inherits but overrides the chi0 computation (`_calc_chi0q_ir`), the vertex assembly (`_flex_compute_veff*`), and the solve loop — and this duplication of the same physics is exactly where this codebase has been accumulating defects.

The proposal, in two strict phases:

1. **Verify equivalence first.** Run RPA and FLEX-with-`IterationMax=1`-and-no-self-energy-feedback on the same inputs and require the outputs to agree to numerical precision, across the full input matrix — not a single smoke case.
2. **Only after that**, make RPA a thin configuration of the FLEX path and retire the duplicate implementation.

## Why

Every one of the following came from two implementations of the same quantity drifting apart, and each cost a full measurement campaign to settle:

- #96 — `sc.py::_build_interaction_k` and `rpa.py::_make_ham_inter` disagreed on the orbital-pair orientation of the same interaction file (fixed in #99).
- #104 — the RPA ring's effective spin vertex is `diag(U0, 0, 0, U1)` where `sc.py`'s S matrix and exact diagonalization both give `diag(U0, U', U', U1)`.
- #106 — `uhfk.py` symmetrises interaction declarations and applies an Ising factor that `rpa.py` does not.
- #90 — RPA's transverse channel carried a vertex table with four of seven entries wrong, undetectable because the channel had no independent reference.
- #91/#78 — the FLEX general path carried a spurious orbital-pair transpose (fixed in #94).

A single implementation makes this whole class of defect structurally impossible for the quantities it covers.

## The equivalence matrix to verify

At minimum, the cross product of:

- `calc_scheme`: `reduced`, `squashed`, `general`
- interaction types: each of the seven, alone and in Kanamori combinations, one- and two-orbital fixtures
- spin modes: spin-free, spin-diag (2Sz constrained), spin-orbital where supported
- chi0q supplied externally (`chi0q_init`) vs computed internally
- outputs: `chi0q`, `chiq`, and every saved artifact byte-for-byte in structure and to `atol=1e-8` in values (matching the repository's test convention)

Known asymmetries to resolve or scope out explicitly, found by inspection:

[...]

Issue #106, opened 2026-07-28
UHFk and RPA read the same interaction file differently: symmetrisation and the Ising factor


k-yoshimi:
## Summary

`uhfk.py` and `rpa.py` read the same interaction file differently. `uhfk.py` symmetrises every two-body type over the two orbital-pair declarations and applies an extra factor for Ising; `rpa.py` does neither. Since the file format lets the same operator be written in more than one way, the two solvers can disagree on what a given input means.

## What each does

`uhfk.py` builds every interaction table as the **average** of the `(a,b)` declaration and its Hermitian partner:

```
CoulombIntra   uab_r                        (on-site same-orbital; symmetric by construction)
CoulombInter   (vab_r + vba)/2
Hund          -(jab_r + jba)/2
Ising          (jab_r + jba)/2 / 4          <- note the extra 1/4
PairLift       (jab_r + jba)/2
Exchange      -(jab_r + jba)/2
PairHop        (jab_r + jba)/2
```

`rpa.py::_append_inter` places each declared entry directly, with a per-type spin table of coefficients (`Ising: {(0,0,0,0): 1, (1,1,1,1): 1, (0,0,1,1): -1, (1,1,0,0): -1}`), with no symmetrisation and no 1/4.

## The consequence, measured

For a density-density interaction `n_a n_b = n_b n_a`, so only `V_ab + V_ba` is physical. An on-site `CoulombInter` with `V_01 = +1` and `V_10 = -1` is therefore an **identically zero** Hamiltonian. Feeding it to the RPA transverse channel:

| input | Hamiltonian | RPA vertex |
|---|---|---|
| `V_01 = +1`, `V_10 = -1` (CoulombInter) | zero | `(0,1),(0,1) = +1.00`, `(1,0),(1,0) = -1.00` |
| the same as Ising | zero | `-1.00`, `+1.00` |

So the RPA response depends on how the coefficient happened to be split between two entries that denote the same operator. `uhfk.py`, which averages, gives zero for the same input.

An asymmetric-but-nonzero declaration such as `V_01 = 1.0`, `V_10 = 0.35` has the same problem in milder form: the two solvers use different effective couplings.

## The open question: sum or average

This has to be settled before either solver is changed, and the available evidence points both ways.

**Average**, per `uhfk.py`: the two entries are a redundant Hermitian declaration of one bond, and the physical coefficient is their mean.

**Sum**, per the operator algebra of Exchange: `X_ab^dagger = X_ba`, so

```
[...]

PR #112, opened 2026-07-29
Accept off-site density-density interactions in the general FLEX path

k-yoshimi:
Closes the off-site feature gap found in the #107 phase-1 verification matrix, where "RPA = FLEX at one iteration" was false at the feature level: the general path rejected every off-site entry while the RPA ring handles off-site V(q) correctly.

## Why the blanket rejection was unnecessary

The guard cited a feared mismatch between the S/C construction's uniform k-grid and chi0's FFT q-grid. Measured, the fear is unfounded — both are the C-ordered `2*pi*i/n` grid. The clean test is **one orbital**, where no inter-orbital slot differences exist to confound the comparison: FLEX at `IterationMax=1` with an off-site CoulombInter reproduces the RPA ring **exactly** (`chi_s` to 5e-15, `chi_c` to 4e-14, on 4x4 and on a non-cubic 4x6 lattice). The density-density family needs nothing beyond the `e^{-iqR}` phases `_build_interaction_k` already carries — the vertex is `V(q)` at the density slots, ordinary extended-Hubbard RPA at one iteration, and the same `V(q)` machinery the reduced path has always used.

## The policy

| off-site entry | verdict | reason |
|---|---|---|
| CoulombInter, Hund, Ising | **accepted** | density-density: vertex is `V(q)`, grids verified to match |
| Exchange, PairHop | rejected | non-local particle-hole pair; no q-only S/C matrix can represent it (the locality argument measured for the transverse channel in #90) |
| CoulombIntra | rejected | UHFk reads only its `r = 0` component; accepting it here would add a cross-solver semantics divergence (#106) |

## Tests

- One-orbital off-site V: exact equivalence with the RPA ring (chi0q, chi_s, chi_c), on 4x4 and non-cubic 4x6.
- Two-orbital off-site V: chi0q exact; the channel difference against the ring is **confined to the inter-orbital pair slots** — the same vertex-content difference as on-site U' (#104), where exact diagonalization already showed the FLEX side correct. Anything outside those slots would be a genuine off-site defect, and the test fails on it.
- Exchange / PairHop / CoulombIntra off-site still rejected (the pre-existing guard test updated from the blanket policy to this one).
- A three-iteration off-site run exercises the q-dependent `V_eff` through the self-energy path and must stay finite.

954 tests pass.


k-yoshimi:
Hardened after the question "is the output completely identical to RPA?" — the honest answer was: not verified to that standard. The one-orbital test compared channel *projections*, silently assuming RPA's chiq obeys uu == dd and has zero spin-mixed pair blocks.

Now element-complete: RPA's full spin-orbital `chiq` is reconstructed from FLEX's channels (same-spin blocks `(chi_c + chi_s)/2`, cross-spin `(chi_c - chi_s)/2`) and **every nd^4 element at every frequency and q** is compared, including the spin-mixed blocks, which must vanish for the reconstruction to be complete. Measured: 2e-14 (4x4), 5e-14 (non-cubic 4x6), spin-mixed blocks exactly zero, and the in-loop chemical potential **bit-for-bit identical** between the two solvers (the `mu` attribute FLEX exposes afterwards belongs to its post-loop dressed bookkeeping and has no RPA counterpart).

What remains outside "complete equality" by design rather than by gap: the on-disk artifact sets differ (RPA saves `chiq`; FLEX saves `chiq_s`/`chiq_c`/`sigma`/`green`/`physics`) — that is the #107 storage-convention question, not a numerical discrepancy.

k-yoshimi:
A strict review round and a parallel measurement converged on the same must_fix: the previous commit's "density-density family" acceptance was a **reasoned class, not a measured one**, and the reasoning was wrong for most of it. Resolved in `9f40f79` by narrowing to exactly what is measured.

**The mechanism (review):** `build_sc_matrices_myo` applies the full on-site Kanamori slot mapping to whatever it receives — including the Fierz (Case 2) inter-orbital slots. Off-site, the particle-hole pair behind those slots is non-local and not representable by a q-only matrix.

**The confirmation (measurement):** off-site Hund and Ising differ from the RPA ring by **3e-2 / 7e-2 even at one orbital**, where no inter-orbital slot exists to blame; and sublattice folding makes even a same-orbital bond differ by 2e-2 (the folded bond becomes intra-cell inter-orbital coupling).

**The accepted class is now exactly the measured-equal one:** CoulombInter with same-orbital pairs (a == b), no sublattice folding — element-complete equal to the ring at 1e-16..1e-14: one orbital on 4x4 / non-cubic 4x6 / 3D 4x4x2 with a z-direction bond; two orbitals on 4x4 / 4x6. For a == b entries nothing reaches the Fierz slots.

**The two-orbital test changed character**, addressing the review's discrimination concern: the masked-confinement assertion (which could not tell off-site contamination from on-site #104 content — same slots) is replaced by an a == b fixture requiring exact agreement in **every** element, no masked region.

Same-orbital off-site V is the extended-Hubbard use case, so the practically important configuration survives. a != b off-site, Hund/Ising off-site and folded off-site wait for #107 or an exact-diagonalization adjudication of the Fierz-slot content.

955 tests pass.

k-yoshimi:
Post-merge confirmation review (with the develop that contains the #113/#114 vertex corrections) found two defects, both fixed in 480b7ef:

- **Folding bypass:** the off-site guard scanned the folded interaction table; a folded dimension of size one canonicalizes every off-site displacement to (0,0,0), so CellShape=[4,4,1] with SubShape=[4,1,1] and ±x bonds slipped past the rejection and diverged. Under folding the guard now scans the pre-fold table. Reproduced before the fix, rejected after.
- **One-sided declarations:** the #114 builders reduce a declaration to its reversal-symmetric part, while the RPA ring kept a one-sided off-site bond as `v e^{-iqR}` — measured 1.2e-2 apart in chiq for the same Hamiltonian. The one-sided declaration means `v n(i) n(i+R)`, whose exact vertex is `v cos(qR)`, so the ring's reading was the wrong one (the RPA side of #106). `rpa.py` now reduces its interaction tables to the same symmetric coefficient as `uhfk.py` and the pairing-vertex builders; one-sided `v` and `v/2` at both ends are now bit-identical, symmetric input is unchanged (the reduction is an identity there — all reference outputs pass), and one-sided off-site input stays accepted with the FLEX/ring equivalence intact.

New tests: the SubShape=[4,1,1] bypass joins the rejection matrix; a one-sided bond is checked element-complete against the ring and bit-for-bit against its both-ends reading. Full suite: 978 passed.

k-yoshimi:
Round-3 review findings, all addressed:

- **Aggregate `Coulomb` was silently dropped (blocker):** the general path knew only the split interaction names, so an aggregate declaration bypassed the off-site guard and produced a zero vertex — measured 1e-1 in chiq_s against the identical explicit declaration, with no warning. The table is now normalized via the shared `split_coulomb` decomposition before both the guard and the vertex build; aggregate and explicit input are bit-identical (03c2c5a), and off-site aggregate entries hit the same per-class policy.
- **Fail-open pre-fold scan:** if folding is active but the pre-fold table is missing, the guard now raises instead of silently validating against the folded table (03c2c5a).
- **Sign-flipped partner lookup out of range / wrapped keys:** already replaced in 0aa9a48 — the reversal is done on a dense grid with the same roll+flip index map `uhfk.py` uses, so wrapped forms like `(n-1,0,0)` and out-of-range forms are canonicalized modulo the lattice; wrapped and signed declarations of one bond verified bit-identical.

Additional tests from the review: one-sided on-site Hund reads bit-identically to its halved both-ends form (the symmetrisation covers every type), aggregate == explicit, off-site aggregate a≠b rejected, missing pre-fold table fails closed. Full suite: 983 passed.

Issue #108, opened 2026-07-29
Extract the computation shared by FLEX and Eliashberg into single-implementation modules


k-yoshimi:
## Proposal

Extract the computation that FLEX and the Eliashberg solver share into modules used by both, so each physical quantity has exactly one implementation.

## Current state

The sharing is partial and, where it exists, takes the form of *parallel implementations kept in sync* rather than one implementation:

- `hwave/solver/_sc_matrices_myo.py` describes itself as *"a sibling of `hwave.sc._build_sc_matrices_all_q` ... so that the two stay element-for-element"* — two builders for the S/C matrices, with tests pinning their agreement, instead of one builder.
- `hwave/sc.py` has its own interaction builder (`_build_interaction_k`), its own Green-function handling, and its own vertex assembly, alongside `rpa.py::_make_ham_inter` and the FLEX vertex code.
- The Eliashberg solver consumes FLEX susceptibilities through files, and the conventions on both sides of that interface have to be maintained by hand (#101 exists because the files do not even record which orientation they were written with).

## Why

The defects this session-long audit surfaced were, almost without exception, disagreements between parallel implementations of the same quantity:

- #96 / #99 — interaction-vertex orientation differed between `sc.py` and `rpa.py`.
- #100 / #102 — PairHop placement had to be settled separately for each builder.
- #95 / #97 — the S/C charge diagonal was missing `2V_aa(q)` in one builder.
- #104 — the ring spin vertex disagrees with the S matrix the Eliashberg side uses.
- #106 — declaration symmetrisation and the Ising factor differ between solvers.

Each was found only by an exact-diagonalization campaign, and each fix then had to be proven not to break the *other* implementation. With one shared module per quantity, agreement is structural and the sync-pinning tests become unnecessary.

## Candidate seams

From the audit, the pieces both sides compute and where divergence has actually occurred:

1. **Interaction reading and vertex building** — one builder from file to four-index tensor, with the symmetrisation convention (#106) implemented once. Consumers: UHFk, RPA/FLEX, Eliashberg.
2. **S/C (spin/charge) matrices** — collapse `_build_sc_matrices_all_q` and `build_sc_matrices_myo` into one.
3. **chi0 bubble** — one kernel (with the IR and dense-Matsubara routes as backends of the same interface, if both are kept).
4. **RPA channel solve** — `[I + chi0 W]^-1 chi0` exists in `rpa.py::_solve_rpa` and in the FLEX vertex path.
[...]

PR #114, opened 2026-07-29
Correct the S/C vertex content for Hund, Exchange and Ising against exact diagonalization

### LINKED ISSUE #113: MYO S/C matrices are wrong for Hund, Exchange and Ising (slot, sign and factor errors against exact diagonalization)
## Summary

Exact diagonalization of the on-site vertices (both spin and charge channels, self-energy-subtracted, mean declaration convention — full methodology and the three-way table in #107) shows that the MYO S/C matrices built by `sc.py::_build_sc_matrices_all_q` / `hwave.solver._sc_matrices_myo` are wrong for the exchange-family interaction types, each in a different way. Two-orbital, J = 0.7, code convention `[I + X0 W]^-1 X0`:

| type | channel | exact (ED) | S/C matrices produce | defect |
|---|---|---|---|---|
| Hund | spin | −J at (a,a)x(b,b) **and +J at (a,b)x(a,b)** | only the −J part | pair-diagonal entry missing |
| Hund | charge | −J and **+J at (a,b)x(a,b)** | −J and **+2J** | factor 2 (possibly a double-count of the two declarations — a single-entry run distinguishes) |
| Exchange | both | ±J at the **pair-diagonal** (a,b)x(a,b) | ±J at the **antidiagonal** (a,b)x(b,a) | wrong slot family |
| Ising | spin | +2J at (a,a)x(b,b) and **−J** at (a,b)x(a,b) | +2J and **+J** | wrong sign on the pair-diagonal entry |
| Ising | charge | −J at (a,b)x(a,b) | −J | correct |

CoulombIntra, CoulombInter, PairLift and PairHop are correct (CoulombInter and PairHop were adjudicated end-to-end earlier in this cycle).

## Why it matters

These matrices are not only the FLEX susceptibility vertices — they feed the Eliashberg pairing vertex. Any calculation whose interaction includes Hund / Exchange / Ising terms (e.g. a full Kanamori set with J) is affected in both the susceptibilities and the superconducting kernel. **U/V-only calculations are unaffected**, which covers the extended-Hubbard charge-fluctuation work.

## Notes

- The same ED table shows the RPA ring is *also* wrong for these types (it simply has no inter-orbital entries at all) — so agreement between the two solvers cannot be used as a check here; the ED table is the target (#107).
- First follow-up measurement: rerun the S/C build with a single declared entry (0.7 one direction only) to separate the Hund-cc factor-2 into "double-count" vs "wrong coefficient".

k-yoshimi:
Closes #113. Methodology and the full three-way (ED / RPA-ring / FLEX) adjudication table are in #107.

## The defects

Exact diagonalization of H-wave's documented file operators showed the S/C interaction matrices were wrong for the exchange-family types, each differently (`chi_s = [1 - chi0 S]^-1 chi0`, `chi_c = [1 + chi0 C]^-1 chi0`):

| type | defect |
|---|---|
| Hund | the spin contribution `S -= J` at (ab,ab) was **missing entirely** |
| Exchange | its ±J sat on the pair-antidiagonal Case-4 slots; the exact vertex puts it on the pair-diagonal Case-2 slots (`S += J'`, `C += J'`) |
| Ising | the spin sign at (ab,ab) was **inverted** (`S += I`, not `S -= I`) |

## The "MYO vs Kuroki convention difference" dissolves

`build_sc_matrices_myo` carried `-U'+2J` at the charge (ab,ab) slot against `sc.py`'s `-U'+J`, documented as a literature-convention choice. The exact per-type attribution is **Hund +J and Exchange +J'**: for the SU(2) Kanamori combination (J = J') their sum reproduces the standard `-U'+2J` exactly, while the spin channel gets `-J + J' = 0` — the standard "no J in S(ab,ab)". So MYO's value was right for the combination but wrong as a per-type split, and Kuroki's was wrong for the combination. Neither was a convention; one was a bookkeeping error and the other an omission.

With the per-type values fixed, the two builders are identical — `build_sc_matrices_myo` now **delegates** to `_build_sc_matrices_all_q` (the first concrete step of #108), and the per-q `_build_sc_matrices` delegates too, retiring two hand-maintained copies (one of which had already drifted once).

## Verification

- Three-way table after the fix: **|ED − FLEX| = 0.000 in all eight cells** (CoulombInter / Hund / Exchange / Ising × both channels), against 0.700 for the RPA ring in every cell — that side is #104, to be resolved by the #107 unification.
- Internal consistency: the SU(2) combination reproduces the standard Kanamori matrices, and the transverse table from #105 is consistent through the SU(2) identity for the same combination.
- The adjudicated table is pinned per type in `tests/test_sc_vertex_adjudicated.py`, including the SU(2)-combination check. Eight existing tests that pinned the old values (including two that asserted the myo–kuroki *divergence*) are updated, with the adjudication recorded at each changed assertion.

## Affected results

Any calculation whose interaction includes Hund, Exchange or Ising file entries — susceptibilities and Eliashberg kernels both. **U/V-only calculations are unchanged** (which covers the extended-Hubbard charge-fluctuation work).

952 tests pass.

### DIFF
--- a/docs/en/source/algorithm/eliashberg.rst
+++ b/docs/en/source/algorithm/eliashberg.rst
@@ -125,16 +125,30 @@ The matrix elements are:
      - :math:`U`
    * - :math:`l_1 = l_3 \neq l_2 = l_4`
      - Cross
-     - :math:`U' - I`
-     - :math:`-U' + J - I`
+     - :math:`U' + I - J + J'`
+     - :math:`-U' - I + J + J'`
    * - :math:`l_1 = l_2 \neq l_3 = l_4`
      - Density
      - :math:`J - 2I`
      - :math:`2U' - J`
    * - :math:`l_1 = l_4 \neq l_2 = l_3`
-     - Exchange
-     - :math:`J' + P`
-     - :math:`J' + P`
+     - Pair hop
+     - :math:`P`
+     - :math:`P`
+
+.. note::
+
+   The Cross-row entries for :math:`J`, :math:`J'` and :math:`I`, and the
+   placement of :math:`J'` in the Cross row rather than the last row, follow
+   the exact-diagonalization adjudication of the interaction vertices: the
+   Hund term contributes :math:`S \mathrel{-}= J`, :math:`C \mathrel{+}= J`,
+   the Exchange term :math:`S \mathrel{+}= J'`, :math:`C \mathrel{+}= J'`,
+   and the Ising term enters :math:`S` with :math:`+I`. For the
+   SU(2)-symmetric Kanamori combination (:math:`J' = J`) these reproduce the
+   standard literature matrices (no :math:`J` in the Cross :math:`S` entry and
+   :math:`-U' + 2J` in the Cross :math:`C` entry). Susceptibility files
+   produced before this correction carry no ``sc_vertex_version`` field and
+   are rejected when the interaction set contains Hund, Exchange or Ising.
 
 The RPA susceptibilities are
 
--- a/docs/en/source/flex/tutorial/tu-index.rst
+++ b/docs/en/source/flex/tutorial/tu-index.rst
@@ -542,7 +542,10 @@ The FLEX solver produces NumPy ``.npz`` files with the following contents:
   to be paired with, and which shape family they have: ``"kuroki"`` for the
   reduced/squashed schemes (spin-orbital shape, ``nd = norb * ns``) or
   ``"myo"`` for the general full-vertex scheme (orbital-pair shape,
-  ``nd = norb^2``, and the MYO value of the ``C(ab,ab)`` charge vertex). The
+  ``nd = norb^2``; the historical MYO-vs-Kuroki difference in the
+  ``C(ab,ab)`` charge vertex was resolved by the exact-diagonalization
+  adjudication of the per-type vertex content, so the two builders now
+  coincide). The
   Eliashberg loader (``hwave_sc``) uses this tag to interpret the orbital
   indices; it is essential for two-orbital systems, where the spin-orbital and
   orbital-pair dimensions coincide (both ``4``) and shape alone is ambiguous.
--- a/docs/ja/source/algorithm/eliashberg.rst
+++ b/docs/ja/source/algorithm/eliashberg.rst
@@ -125,16 +125,29 @@ RPA感受率は
      - :math:`U`
    * - :math:`l_1 = l_3 \neq l_2 = l_4`
      - クロス
-     - :math:`U' - I`
-     - :math:`-U' + J - I`
+     - :math:`U' + I - J + J'`
+     - :math:`-U' - I + J + J'`
    * - :math:`l_1 = l_2 \neq l_3 = l_4`
      - 密度
      - :math:`J - 2I`
      - :math:`2U' - J`
    * - :math:`l_1 = l_4 \neq l_2 = l_3`
-     - 交換
-     - :math:`J' + P`
-     - :math:`J' + P`
+     - ペアホップ
+     - :math:`P`
+     - :math:`P`
+
+.. note::
+
+   クロス行の :math:`J` 、 :math:`J'` 、 :math:`I` の各項、および :math:`J'`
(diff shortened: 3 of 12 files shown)


k-yoshimi:
Review round complete; both must_fix and the should_fix items resolved in `0d57523`.

**must_fix 1 — asymmetric declarations.** The builders now symmetrise every type at entry (plain-transpose mean; conjugated for PairHop, whose complex phase is physical), matching the transverse channel's slot-family rule and UHFk's file semantics. Antisymmetric declarations — identically zero Hamiltonians — now give zero vertices (pinned for four types); asymmetric Exchange gives the mean at both slots; complex Hermitian PairHop keeps its phase.

**must_fix 2 — legacy files.** FLEX outputs now stamp `sc_vertex_version = 2`; both Eliashberg loaders reject an untagged (pre-#113) file when the interaction set contains Hund / Exchange / Ising, with a regeneration message, and accept it for U/V-only input where the content is provably unchanged.

**should_fix — the Exchange slot move is now independently confirmed**, exactly as requested: an end-to-end longitudinal-channel comparison against the exact O(lambda) response (self-energy removed via the HF response), using only the *measured* bubble correspondence and no vertex-side label reasoning:

| placement | residual |
|---|---|
| pair-diagonal (this PR) | **1.6e-3** (the discretisation floor) |
| antidiagonal (old) | 1.0 |

**should_fix — docs**: EN/JA Eliashberg S/C tables now carry the adjudicated Cross-row values and the PairHop-only last row with an adjudication note; the FLEX tutorial and docstrings no longer present the dissolved MYO-vs-Kuroki difference as a live convention. **should_fix — per-q builder** slices to the requested point before delegating (no O(Nq) blowup).

Not yet committed: a repo-resident slow-test port of the ED adjudication machinery itself (it lives in session scripts; the committed tests pin its *results*). Can be added as an env-gated slow test if wanted.

958 tests pass.

k-yoshimi:
Second review round found two defects in the first-round symmetrisation and guard, both fixed in bd97383:

- **Off-site corruption (confirmed by measurement):** averaging with the same-q orbital transpose collapsed a both-ends bond declaration from `v e^{-iqR}` to `v cos(qR)` — the S/C entry vanished outright at q = π/2. The partner of `(R, (a,b))` is `(-R, (b,a))`, i.e. the transpose **at −q**, so the mean is now `0.5 (M(q) + M(−q)ᵀ)` (PairHop keeps the same-q conjugated transpose, which is exact for its Hermitian partner). For both-ends declarations this is an identity; all on-site behaviour is unchanged.
- **Version guard hardening:** `sc_vertex_version` is now decoded and required to equal 2 in both files — version-1, mismatched, or malformed tags are rejected like untagged files. Activation is by interaction content, so an empty Hund/Exchange/Ising block no longer trips the guard.
- Per-q builder: symmetrises on the full grid before slicing, numpy negative-index semantics restored, out-of-range indices raise.

Regression tests added for off-site V(q) preservation at sin(q) ≠ 0, per-q/all-q equality at every point, one-sided declaration means, and the guard's new rejection cases. Full suite: 965 passed.

k-yoshimi:
Third review round: one further guard defect, fixed in 7fad1c7.

- **Version decode was lossy:** `int()` truncated `sc_vertex_version = 2.9` to 2 and accepted it, and non-finite values escaped as an undocumented `OverflowError`. The tag must now decode to a single integral scalar — 2.0 is accepted as 2; 2.9 / inf / complex / non-scalar / non-numeric are rejected as malformed.
- **Cross-solver contract documented:** for complex Hermitian-closed coefficients, UHFk keeps the full complex value (Hermitian mean, summed over both ordered orbital pairs) while the S/C builders fold the effective real part (each slot used once). Both realize the same physical Hamiltonian — the imaginary part of such a declaration is inert, established against exact diagonalization earlier in this series — and the docstring now states the exact relation. For real declarations the conventions coincide identically.
- The momentum-reversal symmetrisation and the per-q delegation were confirmed correct by the review (odd/even grids, no double- or missed-symmetrisation path).

New tests: the version-tag edge cases above, off-site Hermitian-closed PairHop phase preservation, symmetrisation on a mixed odd/even (5,4,3) grid with per-q spot checks, and the full-frequency loader activating the same legacy-file guard. Full suite: 972 passed.

PR #115, opened 2026-07-30
Add the adjudicated inter-orbital content to the RPA ring's spin vertex

### LINKED ISSUE #104: RPA ring drops the inter-orbital term from the spin vertex, disagreeing with the FLEX path and with exact diagonalization

k-yoshimi:
Fixes #104. Second deliverable of #107 phase 1 (the one-shot equivalence matrix, recorded in that issue).

## Problem

The ring's rank-4 interaction tensor carried the density `(aa,bb)` slots of each interaction type but not the `(ab,ab)` pair-cross slots. Its effective spin vertex for on-site inter-orbital input was `diag(U0, 0, 0, U1)` where exact diagonalization and the pairing-vertex builders both give `diag(U0, U', U', U1)` (#104). Measured against the FLEX channel solve at one shot: inter-orbital V off by 1.0e-2, Hund 1.2e-2, U+V+J 3.9e-3 in chiq.

## Fix

The missing spin-resolved content follows from the adjudicated spin/charge table via `W_same = (C−S)/2`, `W_cross = (C+S)/2` per slot, anchored on the slot families the ring already had right. Additions (on-site a≠b only — the adjudicated domain): CoulombInter −U′ same-spin, Hund +J same-spin, Ising −I same-spin, Exchange +J′ cross-spin, at the (ab,ab) slots. The orientation was selected by measurement (element-complete agreement at 3e-16 vs exact reproduction of the old deficiency).

The corrections live in a **separate tensor** added to the longitudinal solve only: the transverse (ladder) assembly keeps reading the base tensor, because the cross-spin Exchange correction would otherwise land in the very block the ladder assembly reads via the crossing relation and double its adjudicated vertex (measured −0.7 → −1.4). reduced/squashed schemes read the density diagonal and are untouched.

## Verified consequences

- RPA general == FLEX one-shot, element-complete, in every adjudicated cell (was 1e-2 off)
- SU(2) restored: chi_zz == chi_pm for on-site CoulombInter under ring+ladder (3e-3 → 4e-34); the expectedFailure sentinel retired per its own instructions
- spin-diag == spin-free bit-for-bit on a spin-symmetric system — one shared W fixes every spin mode
- The independent two-orbital reference model updated with the same content (and its full pair-space bubble; the old palindromic truncation was exact only for the old W support) and agrees with the solver
- reduced/squashed equivalence with FLEX re-verified at 1e-13

## Behavior change

RPA chiq for multi-orbital systems with on-site inter-orbital interactions changes by up to ~1e-2. This is the #104 bug fix; single-orbital and U-only results are unchanged.

Full suite: 987 passed.

### DIFF
--- a/src/hwave/solver/rpa.py
+++ b/src/hwave/solver/rpa.py
@@ -1004,15 +1004,13 @@ class RPA:
             # input (see _make_ham_inter). The correction's (a,b,a,b) slots
             # sit off the 'kaabb' diagonal, so the reduced/squashed reads
             # below are unaffected by construction.
-            ham_long_q = ham_inter_q
-            if getattr(self.ham_info, "ham_fierz_q", None) is not None:
-                ham_long_q = ham_inter_q + self.ham_info.ham_fierz_q
+            fierz_q = getattr(self.ham_info, "ham_fierz_q", None)
+            ham_long_q = (ham_inter_q if fierz_q is None
+                          else ham_inter_q + fierz_q)
             if gpu_active:
                 ham_inter_q = xp.asarray(ham_inter_q)
-                if ham_long_q is not ham_inter_q:
-                    ham_long_q = xp.asarray(ham_long_q)
-                else:
-                    ham_long_q = ham_inter_q
+                ham_long_q = (ham_inter_q if fierz_q is None
+                              else xp.asarray(ham_long_q))
 
             if self.spin_mode == "spinful":
                 chi0q_orig = chi0q
--- a/src/hwave/solver/rpa.py
+++ b/src/hwave/solver/rpa.py
@@ -547,10 +547,30 @@ class Interaction:
             # with a != b: that is the adjudicated domain -- off-site
             # cross-orbital content is not representable by a q-only vertex
             # (#105 measurement) and stays as it was.
+            # `tbl`, when given, must be a PRE-fold table. Locality is
+            # judged BEFORE sublattice folding: folding maps an off-site
+            # bond to (0,0,0) between supercell orbitals, which would
+            # smuggle unadjudicated off-site content into the correction
+            # (measured: a one-orbital +-x bond folded with SubShape=[2,1,1]
+            # produced a spurious Fierz tensor of magnitude V). The pre-fold
+            # table is filtered to its on-site a != b entries, and only the
+            # filtered table is folded.
+            has_sub = getattr(self.lattice, "has_sublattice", False)
             if tbl is None:
-                tbl = self.param_ham[type]
-            tbl = _symmetrised(type, tbl)
+                if has_sub:
+                    tbl = self.param_ham_orig.get(type, {})
+                else:
+                    tbl = self.param_ham.get(type, {})
+            filtered = {}
             for (irvec, orbvec), v in tbl.items():
+                if tuple(irvec) == (0, 0, 0) and orbvec[0] != orbvec[1]:
+                    filtered[(irvec, orbvec)] = v
+            if not filtered:
+                return
+            if has_sub:
+                filtered = self._reshape_interaction(filtered, False)
+            filtered = _symmetrised(type, filtered)
+            for (irvec, orbvec), v in filtered.items():
                 if tuple(irvec) != (0, 0, 0):
                     continue
                 a, b = orbvec
@@ -583,7 +603,12 @@ class Interaction:
 
             _append_inter('CoulombIntra', coulomb_intra)
             _append_inter('CoulombInter', coulomb_inter)
-            _append_inter_cross('CoulombInter', -1.0, { (0,0,0,0): 1, (1,1,1,1): 1 }, tbl=coulomb_inter)
+            _append_inter_cross(
+                'CoulombInter', -1.0, { (0,0,0,0): 1, (1,1,1,1): 1 },
+                tbl=(wan90.split_coulomb(
+                        self.param_ham_orig['Coulomb'])[1]
+                     if getattr(self.lattice, "has_sublattice", False)
+                     else coulomb_inter))
             self._has_interaction = True
             self._has_interaction_coulomb = True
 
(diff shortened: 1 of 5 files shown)


k-yoshimi:
First review round found one blocker, fixed in b2e321b:

- **Folding bypass in the Fierz domain filter:** the on-site filter ran on the folded table, and folding maps an off-site bond to (0,0,0) between supercell orbitals — a one-orbital ±x bond folded with SubShape=[2,1,1] produced a spurious Fierz tensor of magnitude V (reproduced, then verified fixed). Locality is now judged on the pre-fold table (with the aggregate `Coulomb` split before filtering), and only the filtered on-site a≠b entries are folded.
- The GPU double-upload noted by the review was already fixed in b4fcb05.
- Test hygiene: the spin-mode test uses a private temp directory; the aggregate/explicit equality test now also compares the Fierz tensor; new regression tests cover folded off-site (no-op), folded on-site inter-orbital (correction survives per subcell), one-orbital/U-only (no tensor), and aggregate == explicit under folding.

Review's consumer audit confirmed: FLEX's reads do not need the correction (density-diagonal projection / its own S/C build), the transverse assembly correctly keeps the base tensor, and all three spin-mode branches receive the corrected one. Full suite: 991 passed.

k-yoshimi:
Second review round: no defects found — the round-1 fixes were confirmed correct and complete (pre-fold filtering, aggregate pre-fold split, folding-helper purity, spin-orbital stride, GPU aliasing, all verified by the review). Two of its recommended tests added in eaaa2a0: the folded off-site aggregate bond is asserted to be a no-op, and the Fierz tensor's support is pinned slot-exactly (the four same-spin pair-diagonal entries at −V, nothing else). Full suite: 993 passed.

k-yoshimi:
Final review pass over the whole diff: no defects — judged merge-ready. The physics content was independently re-verified (per-type coefficients against the channel decomposition, pair-diagonal on-site a≠b placement, no correction for CoulombIntra/PairLift/PairHop, longitudinal-only consumption, reference-model construction, and that the inverted SU(2) tests assert full-matrix equality). Its remaining recommendation — slot-exact support assertions for all four corrected types, not just CoulombInter — added in d10db4c. Full suite: 993 passed.

PR #116, opened 2026-07-30
Validate an in-memory chi0q and set the spin mode before use

### LINKED ISSUE #109: RPA.solve crashes with AttributeError (spin_mode) when green_info carries a chi0q from a previous solve
## Symptom

Calling `RPA.solve()` a second time with a `green_info` that carries the `chi0q` stored by a previous solve crashes:

```
File "src/hwave/solver/rpa.py", line 914, in _solve_impl
    if self.spin_mode == "spinful":
AttributeError: 'RPA' object has no attribute 'spin_mode'
```

Reproduction (library usage, no files involved beyond the standard 2-orbital fixture):

```python
solver1.solve(green_info, out)      # ring, computes green internally; stores chi0q
solver2.solve(green_info, out)      # fresh solver, same green_info -> AttributeError
```

After the first solve, `green_info` contains `['chi0q', 'chiq']`. The second solver sees `chi0q` present, takes the externally-supplied-chi0q branch, and reaches the `spin_mode` dispatch at `_solve_impl` (line 914) before anything has set `spin_mode` on this instance.

## Notes

- Found while writing a state-hygiene test for #105; the test had to sidestep this path (it injects stale keys directly instead of reusing the full dict) and documents the sidestep.
- The stale `chiq`/`chiq_pm` half of the reuse problem was fixed in #105; this crash is the remaining half.
- Whether `chi0q_init` supplied via the input file enters through a path that does set `spin_mode` has not been checked; the crash above is the in-memory reuse path.
- Related in spirit to #107: if RPA becomes a configuration of the FLEX path, this dispatch is one of the code paths that would be unified away.

## Suggested fix

Set `spin_mode` during the chi0q shape inspection unconditionally before any consumer, or fail with a clear message when an externally-supplied chi0q reaches a solver whose spin structure was never established.

k-yoshimi:
Fixes #109.

Handing a `green_info` stored by a previous solve to a fresh solver — ordinary library usage, e.g. scanning interaction parameters on a fixed bubble — crashed with `AttributeError: 'RPA' object has no attribute 'spin_mode'`: the externally-supplied-chi0q branch reached the spin-mode dispatch with `spin_mode` never set. The file-based `chi0q_init` route ran the shape detection inside `_read_chi0q`; the in-memory route ran none.

The shape validation and spin-mode detection now live in one shared method called by both routes. Junk shapes on the in-memory route fail fast at the validation with a clear message instead of deep inside the spin inflation.

Tests (written first, reproducing the crash): the green_info round-trips through a second solver for all three calc_schemes and reproduces the first solver's chiq exactly; a two-block bubble handed over in memory selects spin-diag mode and matches the spin-free result on symmetric input.

Full suite: 997 passed.

### DIFF
--- a/src/hwave/solver/rpa.py
+++ b/src/hwave/solver/rpa.py
@@ -959,8 +959,13 @@ class RPA:
             # as the file-based chi0q_init route does (issue #109)
             chi0q = green_info["chi0q"]
             self._validate_chi0q_shape(chi0q, source="green_info")
-            if chi0q.shape[0] != self.nmat:
-                logger.info("partial range in matsubara frequency: {} in {}".format(chi0q.shape[0], self.nmat))
+            # spin-diag arrays carry the spin-block axis first; the
+            # frequency axis is shape[1] there (shape[0] == 2 == nblock,
+            # which used to be misreported as a partial frequency range)
+            nfreq = (chi0q.shape[1] if self.spin_mode == "spin-diag"
+                     else chi0q.shape[0])
+            if nfreq != self.nmat:
+                logger.info("partial range in matsubara frequency: {} in {}".format(nfreq, self.nmat))
                 #self.nmat = chi0q.shape[0]
             if gpu_active:
                 # VRAM preflight for the externally-supplied chi0q path: the
@@ -1412,12 +1417,27 @@ class RPA:
                 # spin-free or spinful
                 #   shape = (nmat,nvol,nd,nd,nd,nd) where nd = norb or norb*nspin
                 cs = chi0q.shape
-                assert cs[1] == self.lattice.nvol, "lattice volume"
+                if not (cs[1] == self.lattice.nvol):
+                    raise ValueError(
+                        "chi0q from {}: lattice volume (shape {})".format(
+                            source, chi0q.shape))
                 nd = cs[2]
-                assert nd == self.nd or nd == self.norb, "shape[2]"
-                assert cs[3] == nd, "shape[3]"
-                assert cs[4] == nd, "shape[4]"
-                assert cs[5] == nd, "shape[5]"
+                if not (nd == self.nd or nd == self.norb):
+                    raise ValueError(
+                        "chi0q from {}: shape[2] (shape {})".format(
+                            source, chi0q.shape))
+                if not (cs[3] == nd):
+                    raise ValueError(
+                        "chi0q from {}: shape[3] (shape {})".format(
+                            source, chi0q.shape))
+                if not (cs[4] == nd):
+                    raise ValueError(
+                        "chi0q from {}: shape[4] (shape {})".format(
+                            source, chi0q.shape))
+                if not (cs[5] == nd):
+                    raise ValueError(
+                        "chi0q from {}: shape[5] (shape {})".format(
+                            source, chi0q.shape))
 
                 if nd == self.nd:
                     self.spin_mode = "spinful"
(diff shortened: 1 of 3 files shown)


k-yoshimi:
Review round 1 found one blocker and two smaller issues, all fixed in 9e0ee5c:

- **Bare `assert` validation (blocker):** under `python -O` malformed input was silently accepted and assigned a spin mode (reproduced by the review with shape (3, 99, 7, 8), and now covered by a subprocess regression test). Every check raises `ValueError` carrying the source and offending shape; the file route's failure type is standardized accordingly.
- **Wrong frequency axis in the partial-range check:** spin-diag input carries the spin-block axis first, so a full-range two-block bubble was misreported as `partial range 2 in 32`; the check now reads the axis selected by the established spin mode, with a log-capture test.
- **File-route log stability:** the validator's new INFO line is demoted to debug, so `chi0q_init` runs log exactly as before.

Also added the recommended spinful-dispatch test for the in-memory route. Full suite: 1000 passed.

k-yoshimi:
Second review round: no must_fix — the round-1 conversions were confirmed correct (no inverted or dropped condition, correct frequency-axis selection, file-route logs unchanged). Its one should_fix is addressed in 8d51e14: `_read_chi0q` now documents `ValueError` as the deliberate public failure type (the previous bare `AssertionError` was disabled by `python -O`), and new tests pin the file route's exception type and message through `read_init` plus the genuinely-partial spin-diag case reporting the frequency-axis length. Full suite: 1002 passed.

k-yoshimi:
An adversarial third review round (run before merge on request) surfaced a cluster of defects on the reuse path, all newly reachable once the crash was fixed, all addressed across 44e30f7 and 4ae0d5d:

- **Frequency-axis mislabeling** (found independently by my own probe and the review): a partial-range bubble reused under a full-range config was saved with a fabricated full axis (`freq_index 0..31` on 9-frequency data). The producing run's metadata now travels with the green_info and is inherited exactly as the file route's `_chi0q_init_meta`; untagged partial arrays get the explicit ambiguous labeling.
- **Stale provenance**: replaced-bubble + leftover metadata is now validated (axis length/uniqueness/range, producer norb/scheme/spin mode) and rejected loudly.
- **Spin-mode collision**: a spin-free two-orbital bubble is shape-identical to a spinful one-orbital one; the recorded producer identity now catches the cross-configuration reuse.
- **Stale results**: chiq/chiq_pm from a previous solve are dropped at solve entry, unconditionally — a failed validation or chi0q-only run can no longer leave them to be saved as this run's output.
- **Ladder hazard**: the spin-diag transverse channel rejects externally supplied bubbles (a same-instance green0 may belong to an older one).
- Device arrays normalized to host when GPU is off; non-numeric dtypes rejected; docstring contract scoped.

Six new regression tests cover each item; the partial spin-diag solve is now asserted to complete rather than allowed to fail. Full suite: 1009 passed.

k-yoshimi:
Round 4 (confirmation of the round-3 hardening) found a further set of defects, all fixed in f648cf8:

- the ambiguous fallback label counted the spin-block axis of a spin-diag bubble (16-frequency data labeled `[0, 1]`) — the frequency axis is counted now;
- the dtype probe broke on real device arrays before the intended host normalization — reordered;
- exact `calc_scheme` equality rejected a legitimate reduced→squashed reuse (verified equal to an internal squashed recomputation, max error 0.0) — only the bubble representation class must agree now;
- the external spin-diag ladder rejection now fires before the longitudinal solve populates chiq;
- an internal recomputation clears stale `chi0q_init` metadata on the same instance;
- provenance schema validation (1-D integer freq_index, positive nmat, empty-safe extrema) and file-route dtype rejection.

Five new regression tests, one per behavior. Full suite: 1014 passed.

k-yoshimi:
Round 5 (convergence check) found the provenance handling still bypassable in two ways, both reproduced by the review and fixed in d0fa1ec:

- **Conditional schema validation:** `nmat` was checked only when `freq_index` was present, fractional values were written back unnormalized, non-mapping metadata raised `AttributeError`, and the `chi0q_init` file route recorded provenance unvalidated. One shared validator now normalizes and checks the mapping on both routes.
- **Same-shaped replacement:** shape checks cannot detect a replaced array of identical shape (a zero bubble inherited the displaced array's `[12..20]` axis). The provenance now carries a content fingerprint (strided blake2b sample — negligible cost, equal for faithful copies) and a mismatch is rejected as stale.

Review-recommended test strengthening included: post-exception state assertions on the ladder rejection, metadata-level scheme rejection isolated from the shape check, schema tables (non-mapping / orphan nmat / fractional / boolean), faithful-copy acceptance, and file-route provenance validation. Full suite: 1019 passed.

k-yoshimi:
Round 6 found three remaining provenance bypasses, all reproduced and fixed in c8f8170:

- **nmat strictness:** size-1 coercion accepted `[32]`, `'32'`, `32.0`, `32+0j` and leaked `TypeError` on mappings. `operator.index` semantics now apply (npz 0-d integer scalars accepted; everything else is `ValueError`), and any `collections.abc.Mapping` is accepted as provenance.
- **Fingerprint completeness:** the strided sample missed single-element edits at unsampled positions — it did not truly bind metadata to the tensor. The full canonical byte content is hashed now (linear cost, small next to the solve).
- **File-route binding:** `chi0q_init` metadata is now bound to the loaded tensor's fingerprint, so replacing `info['chi0q']` between `read_init` and `solve` is rejected instead of inheriting the file's axis.

Regression tests for each (including the honest lesson that the solver's chi0q can be F-ordered, where `reshape(-1)` edits are silently lost — the test writes through `.flat`). Full suite: 1023 passed.

k-yoshimi:
Round 7 narrowed to a single one-line residue of the round-6 fix — the fingerprint fetch was dict-only while the validator accepts any Mapping, so a UserDict's stale fingerprint went unverified. Fixed in 8334ee0 with the review's three recommended tests (UserDict fingerprint verification, legitimate nmat forms accepted, legacy-file end-to-end labeling). Everything else in scope was verified clean by the review: strict nmat probes, full-content hashing across memory orders, file-route re-fingerprinting, save_results isolation. Full suite: 1026 passed.

PR #118, opened 2026-07-30
Single-source the adjudicated vertex content

k-yoshimi:
First increment of #107 phase 2 (design recorded in that issue).

The adjudicated per-type S/C vertex values existed in two implementations — the pair-space case placements in `hwave.sc` and the spin-resolved Fierz coefficients in `rpa.py` — tied only by a derivation recorded in a commit message. That duplication pattern is exactly how #96, #104 and #113 happened.

One module (`hwave.solver.vertex_table`) now holds the table with its exact-diagonalization provenance, and both builders derive from it; the channel decomposition `W_same = (C−S)/2`, `W_cross = (C+S)/2` is executed in code.

Zero behavior change, verified by the full net built this week: adjudicated-table tests, the RPA==FLEX one-shot equivalence matrix, the SCF fixed-point matrix (18 cells), the transverse suite, and the independent two-orbital reference. A new test measures the decomposition identity from the assembled objects, so an edit bypassing the table on either side fails regardless.

Full suite: 1031 passed.

### DIFF
--- a/src/hwave/sc.py
+++ b/src/hwave/sc.py
@@ -17,6 +17,8 @@ import sys
 import logging
 
 import numpy as np
+
+from hwave.solver.vertex_table import sc_coefficients
 from numpy.fft import fftn, ifftn
 from scipy.optimize import bisect
 from scipy.sparse.linalg import LinearOperator, eigs, bicgstab, gmres, lgmres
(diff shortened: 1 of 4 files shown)


k-yoshimi:
Review round 1: one behavioral edge (zero coefficients multiplied instead of suppressed — `0.0 * Inf = NaN` contaminated a channel that previously stayed exactly zero; reproduced by the review) and one docstring contradiction (the deliberate a == b application of CoulombInter's density entry), both fixed in 273a114 with the review's two recommended tests (direct table pin, non-finite-input suppression). Everything else was verified clean: all seven table entries against the adjudicated tests, all four case families, the four derived ring mappings, the aggregate-Coulomb call site, and the consistency test's indexing. Full suite: 1033 passed.

k-yoshimi:
Round 2 found one remaining IEEE edge — `1.0 * (Inf+0j)` is `Inf+NaNj` under numpy's complex multiply, so ±1 coefficients must use direct add/subtract to keep the pre-refactor component structure for non-finite couplings. Fixed in 8fb2fab with a coefficient-aware accumulator (zero suppression / direct ±1 / multiply otherwise) across all four case families, the docstring claim narrowed to the actually single-sourced pieces, warnings silenced in the non-finite tests, and an isolated IEEE parity test (fed pre-symmetrised arrays, since the public Fourier builder itself converts Inf to Inf+NaNj identically in old and new code). Full suite: 1034 passed.

k-yoshimi:
Round 3: **converged** — no must_fix, no should_fix. The review verified the accumulator's write-through on sliced views (runtime-checked), the unchanged ±2 path, the narrowed docstring, and the test isolation. Its optional items (scoped warning filters via catch_warnings, a Case-3 Hund density parity case, an explicit ±2 multiply assertion) added in 9f96a6e. Full suite: 1035 passed.

k-yoshimi:
An adversarial fourth round (requested before merge) found one blocker — the exported table was publicly mutable, so a consumer mutating one nested row would silently corrupt both builders (reproduced by the review). Fixed in 2628e9c: both mapping levels are read-only proxies with no mutable backing name retained.

Its test-hardening recommendations are all in: the ±2 form pinned by observable legacy behavior (`2.0 * (Inf+0j)` → `Inf+NaNj` under numpy's promotion — identical to the pre-refactor multiply, distinguishable from a two-adds replacement), the decomposition test now asserts the expected nonzero per-type values before the identity (no vacuous pass), and a permanent mixed-type norb=3 golden test freezes selected slots of every type and family after the review's independent bit-for-bit old/new parity probe.

The round's adversarial probes verified: bit-for-bit sc parity for norb 1–4 / all types / mixed / complex / non-finite, exact ring Fierz tensors incl. aggregate Coulomb, in-place view semantics, and no import cycles. Full suite: 1036 passed.

k-yoshimi:
Round 5: **converged** — no must_fix, no should_fix. The review verified the freeze end to end (both levels mappingproxy, no mutable backing name, immutable returns, mutation attempts fail) and confirmed the ±2 expectations against numpy's actual promotion behavior. Its optional items (a direct freeze-pinning test, a corrected docstring) added in 440ab76. Full suite: 1037 passed.

PR #119, opened 2026-07-30
Single-source the reduced/squashed density projections

k-yoshimi:
Second increment of #107 phase 2 (plan in that issue; increment 1 was #118).

The density-diagonal reduction of the interaction tensor — the defining operation of the reduced and squashed schemes — existed as seven inline einsum expressions across the two solvers, in two index notations. They now live once in `hwave.solver.density_projection`. FLEX's spin-factorized `'ksasatbtb->ksatb'` is the same extraction as the combined-index `'kaabb->kab'` (verified bitwise on random complex tensors), so all plain-reduced sites share one function; the squashed form keeps its own.

Zero behavior change, pinned by the RPA==FLEX one-shot equivalence matrix, the SCF fixed-point matrix (18 cells), and the full suite: 1037 passed.

### DIFF
--- a/src/hwave/solver/density_projection.py
+++ b/src/hwave/solver/density_projection.py
@@ -1,0 +1,41 @@
+"""Density-diagonal projections of the rank-4 interaction tensor.
+
+The ``reduced`` and ``squashed`` calculation schemes solve with the
+density-density diagonal of the interaction: exchange-, spin-flip- and
+pair-type off-diagonal vertices are dropped (a deliberate, documented
+approximation -- the callers warn about it). The two solvers used to
+carry their own copies of these einsum reductions; per issue #107 the
+projections live here once.
+
+Layout conventions:
+
+* ``project_density_pairs``: combined-index form. The tensor is read as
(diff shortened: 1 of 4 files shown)


k-yoshimi:
Review round 1: no must_fix; its one should_fix was real — the sweep had missed an eighth site (the RPA spin-free/reduced branch's inline projection), now routed with a grep-verified zero remainder. The recommended unit tests pin both helpers exactly against the original einsum notations, including the spin-major layout identity. The review's own equivalence audit confirmed all routed sites preserve their original dimensions and output layouts. Full suite: 1040 passed.

k-yoshimi:
An adversarial third round (requested before merge) found no behavioral defect — its probes independently verified all eight sites, dimension derivations under unusual configurations (spin-orbital, folding, partial frequency axes), GPU dispatch, absence of downstream in-place mutation, and view/copy equivalence across C/Fortran/non-contiguous layouts. Its one should_fix (module doc implied the squashed helper serves every squashed scheme; FLEX deliberately routes both its modes through the pair projection) and the layout/degenerate-shape test matrix are addressed in b84cf38. Verdict quoted: 'Behavioral implementation is converged.' Full suite: 1041 passed.

PR #120, opened 2026-07-30
One Exchange/PairHop policy for the density-diagonal schemes

k-yoshimi:
Third increment of #107 phase 2 (plan in that issue).

## Problem

Exchange and PairHop have no density-diagonal vertex content (adjudicated: their entries live on the cross/antidiagonal slot families), so the reduced and squashed schemes were not approximating them — they dropped them entirely, with exactly zero effect. Three policies coexisted: RPA reduced rejected exchange-type input, squashed accepted it silently, FLEX warned of an 'approximation'.

## Change (behavioral, deliberate)

- reduced/squashed **reject** nonempty Exchange or PairHop input (ValueError pointing to `calc_scheme='general'`), both solvers, one shared check.
- PairLift stays accepted everywhere (its vertex is exactly zero — dropping it is exact); the warning now says that instead of 'approximated'.
- `calc_scheme='auto'` selects **general** when Exchange/PairHop content is present (the historical choice was squashed — silently dropping them; PairHop-only input even fell through to reduced).

## Impact

Runs that previously used squashed/auto with Exchange or PairHop were producing results in which those interactions did nothing; they now either error with guidance (explicit squashed) or get the correct full vertex (auto). The iron-pnictide Kanamori benchmark runs under the full adjudicated vertex with Exchange genuinely acting for the first time — all six physics claims still hold. Policy pins updated across the fixed-point matrix and scheme-guard tests.

Full suite: 1041 passed.


k-yoshimi:
Review round 1 caught the shipped ecosystem still depending on the old policy — all addressed in b985574:

- **Sample scripts (blocker):** the iron-pnictide sample's `auto` now resolves to general, whose rank-6 chiq crashed the reduced-layout plotting scripts (three copies: docs/en, docs/ja, docs/tutorial). All three use the layout-aware density-sector extraction now.
- **Tutorials (blocker):** the EN and JA texts stated the old 'partially supported with a warning' behavior in four places each; they now state the rejection policy, auto→general, PairLift's exact-zero acceptance, and the compatibility consequences. The stale 'general is on-site only' claim (outdated since the off-site same-orbital class landed) is corrected in passing, and Sample 3b's text explains that Sample 3's auto now runs the identical full-vertex path.
- **Guidance (should_fix):** the rejection message notes that no current FLEX scheme carries these interactions under enable_spin_orbital / spin-polarized setups, instead of pointing at another rejected configuration.
- Stale test docstrings rewritten; direct-RPA rejection pins, the auto-selection matrix, and an explicit benchmark scheme assertion added per the recommendations.

Full suite: 1044 passed.

k-yoshimi:
Round 2 caught that the reduced-layout fallback in the plotting scripts summed only the density-matrix diagonal — the physically consistent contraction sums the whole density-pair matrix, as the general branch and the benchmark already do. Fixed in 680a142 across all five copies (including the two general-scheme sample scripts the faulty branch had been ported from), plus the two stale off-site comments qualified to the actual acceptance class. Full suite: 1044 passed.

k-yoshimi:
Round 3 found a sixth tracked copy of the plotting script (docs/tutorial's general variant) still on the diagonal-only contraction — fixed in 6e00e55, with a repository-wide grep confirming no remaining copy. The round verified the five previous fixes (correct (nvol, norb, norb) slicing and axis handling) and both qualified comments against the implemented policy.

Issue #92, opened 2026-07-28
Eliashberg uses only the first spin block of a two-block dressed green.npz


k-yoshimi:
Split out of #87, where a guard for this was written, revised three times, and then removed as out of scope. Recording it so the gap is not lost.

## The gap

`_load_flex_green` takes `green_raw[0]` unconditionally. A `spin_mode = "spin-diag"` FLEX run writes **two** blocks, and on a two-orbital Zeeman fixture they differ by 5.59 — so the down-spin propagator is discarded, and the pair bubble built from the remaining one feeds the Eliashberg kernel directly.

## Why it is not urgent

The susceptibility guard added in #87 rejects a spin-polarized run before this function is reached, so the realistic path is covered. What remains is the case where the susceptibility happens to be redundant while the Green functions are not — contrived, but not impossible for an externally assembled input.

## Why the guard was removed rather than kept

It is a different quantity from the one #87 is about, and it accumulated its own defects across three review rounds:

1. exact equality where the susceptibility check allowed round-off, so a last-bit difference rejected a valid producer;
2. `max(gscale, 1.0)` instead of a relative test, which for small amplitudes accepted far more than intended — and the test written to cover it passed only because its fixture amplitude was 0.01;
3. the same global-ratio masking defect later found and fixed in the susceptibility guard: `max diff over all frequencies / max scale over all frequencies` lets a large redundant frequency hide a small completely non-redundant one (frequency 0 at 1e14 both blocks, frequency 1 at 1 and 0 gives ratio 1e-14, under the threshold).

Carrying a guard that needed that many corrections inside a bug-fix PR was making the change riskier than the bug it protected against.

## If it is picked up

Take the susceptibility guard as the template — it has been through the same failures:

- ratio per frequency against that frequency's own scale, not a global pair;
- reject non-finite values explicitly, since `max(0.0, nan)` is `0.0` in Python and a NaN would otherwise leave the running maxima untouched;
- accumulate per frequency rather than allocating whole-array temporaries (a production Green function is multi-GB);
- regression with a large redundant frequency beside a small non-redundant one, verified to fail against the unfixed version rather than merely passing.

k-yoshimi:
[...]

PR #121, opened 2026-07-30
Embed reduced-scheme susceptibilities as density pairs in the Eliashberg vertex

k-yoshimi:
Lands the fix from #87 onto the current `develop` (the original branch predated the #94–#120 series and had gone stale; this branch merges it and resolves the conflicts, so the original commits and authorship are preserved in the history).

## Problem (from #87, still present on develop — reproduced before this merge)

A reduced/squashed run stores its susceptibility with a **density-pair** matrix index: `X[a,b]` *is* χ_{(a,a),(b,b)}. Two Eliashberg-side loaders lifted it into orbital-pair space by reading that index as an ordinary orbital index (a `kron(X, I)` scatter), which (1) dropped the genuine inter-orbital density coupling χ_{(0,0),(1,1)} that `S χ S` references, and (2) fabricated dressing at pair indices the reduced scheme never computed. Both the static and dynamic `chi0q_mode="flex"` routes and the `"load"` route were affected; multi-orbital reduced-scheme λ and gap functions change (up to ~40 % in the original measurements) and need recomputation. norb = 1 and general-scheme results are bit-identical.

## Additions beyond the core fix (from #87)

- A missing-component warning when a reduced susceptibility is combined with interactions whose S/C matrices have off-density weight (decided on the assembled matrices, so cancelling combinations stay silent).
- A guard for spin-diag two-block `green.npz` files (only the up block was consumed silently — the loader half of #92), with a documented `accept_up_block_only` escape hatch and file tags.
- EN/JA documentation warnings describing the result change, and a dedicated test suite (Σ=0 equality across the three routes, embedding unit pins, warning semantics).

## Adaptations made while landing

- The cancellation example updated to the adjudicated slot table (#113 moved Exchange out of the antidiagonal family; equal CoulombInter+Hund now provides the exact cross-slot cancellation).
- The branch's 'general accepts on-site terms only' claims (warning text, EN/JA notes) updated to the actual off-site acceptance class from #112.
- Call signatures reconciled with the #114 version guard (`interactions=` threading) and the round-hardened loaders.

Full suite: 1097 passed (51 new tests from the landed suite).

Supersedes #87.


k-yoshimi:
Review round 1 found one blocker and three documentation drifts, all addressed in c2b8886:

- **Green two-block guard (blocker):** the guard was described but not implemented — the loader unconditionally took block 0 (the original branch had knowingly deferred this; the PR text overstated it). Implemented now, mirroring the chi-side policy: differing blocks raise with guidance, `accept_up_block_only` authorizes with a warning, NaN-identical blocks count as redundant, both uniform and IR routes covered, with loader-level tests for each case.
- The consolidated EN/JA scheme tables updated from 'general is on-site only' to the actual off-site acceptance class; the warning-machinery comments updated to the adjudicated (#113) slot memberships with the correct cancellation example; the tolerance wording now states the actual `256*eps` threshold; trailing whitespace removed.

The round verified the conflict resolution end to end (call arities, `interactions=` threading, version guard, legacy-tag flow, both embeddings, MYO passthrough, provenance isolation). Full suite: 1103 passed.

k-yoshimi:
Round 2 found two sharp defects in the round-1 guard, both reproduced by the review and fixed in 3216dbe:

- **Complex equal_nan semantics:** numpy's `array_equal(equal_nan=True)` on complex arrays treats an entry as equal when *either* component is NaN — `nan+1j` silently equalled `nan+2j`. Components are compared separately now (exact semantics: matching NaN positions, equality elsewhere), with rejection tests for both component orders.
- **Memory behavior:** the whole-block comparison allocated masks comparable to a multi-GB Green block; the scan is per-frequency with early exit, bounding temporaries to one (nvol, norb, norb) slice.
- The `accept_up_block_only` green-side authorization (an uncontrolled approximation, unlike the legacy-layout case) is documented explicitly in EN and JA, and the threshold wording is exact.

The round confirmed everything else: guard placement on both routes, logger/level, test coverage, and the corrected tables. Full suite: 1105 passed.

k-yoshimi:
CI caught a defective fixture in my round-2 tests (my local gate was mis-chained and let a failing test through — the CI failure was the correct signal): the 'NaN-identical' blocks were built with `* 1.0`, a complex multiply that turns `nan+0j` into `nan+nanj` — which under the exact per-component semantics is genuinely differing content, so the guard rightly rejected it. The fixture copies now; fixed in 7546891. Full suite: 1103 passed locally.

k-yoshimi:
Round 3: **converged** — no must_fix, no should_fix. The review verified the per-component comparison, the view/no-copy and short-circuit properties, both routes, the fixed fixture, and the docs against behavior. Its two recommended tests (instrumented slice-granularity/short-circuit pin, guidance-text assertions on the complex-NaN rejections) added in 91aafaa. Full suite: 1104 passed.

k-yoshimi:
An adversarial fourth round (requested before merge) found two real blockers, both fixed in 1ec2e52:

- **Policy inconsistency with #120:** the Eliashberg side still treated Exchange/PairHop + reduced susceptibility as 'approximate and warned' — but those types have zero density-diagonal vertex content (nothing is dressed), and since #120 no reduced/squashed producer can even create such a run: the input is necessarily stale or interaction-mismatched. The combination is rejected now with that reasoning; the warning is reserved for the genuinely partial types, and its per-term attribution goes through the symmetrised builders (an inert declaration is no longer named as a cause). EN/JA docs, test narratives and the Kanamori fixtures updated accordingly.
- **Green loader shape validation:** empty frequency axes (bogus zero kernel), aliased element-count reshapes, and zero-block IndexErrors — all reproduced by the review — are rejected with exact dimension checks before the block comparison.

Full suite: 1105 passed.

k-yoshimi:
Round 5 found the last inconsistencies, fixed in f57f79d: the detailed EN/JA notes and the helper docstring now distinguish the warned partial types from the rejected Exchange/PairHop (matching the summary tables), the rejection message gives producer-neutral recovery guidance, and the two Kanamori tests recover their promised J' coverage by embedding their density chi0q into the four-index form — which also proves four-index input is not falsely rejected. Full suite: 1105 passed.

k-yoshimi:
Round 6: two of the three must_fix items were real and are fixed in d5563b9 — the iteration test now asserts convergence before its eigenvalue comparison, and the rejection message/docstring no longer assert staleness for arbitrary external producers (reduced data simply cannot represent these interactions; the general four-index susceptibility is the universal remedy). Both should_fix items (shared embedding helper at all three sites, the singlet/triplet docstring crediting the full Kanamori set) are in.

The remaining must_fix — S/C[(1,2)] = 0.5 with Jp = 0.5 — is an environmental false positive: those assertions match the adjudicated #113 slot table (Exchange contributes at the pair-diagonal cross slots, not the antidiagonal), and the focused test class passes under the branch source with an explicit PYTHONPATH. The review sandbox imported the editable-installed package from the main checkout — a known trap of this repository's worktrees. Full suite: 1105 passed.

k-yoshimi:
A deep seventh round found two genuine remaining defects, both reproduced and fixed in cf3994c:

- **norb = 1 bypass:** the one-orbital shortcut returned before the Exchange/PairHop rejection, so single-orbital reduced input silently omitted those interactions on both routes (zero vertices measured). The rejection now precedes the shortcut, which only skips the partial-dressing warning.
- **Non-finite greens:** NaN-identical blocks were deliberately accepted, but nothing downstream validates finiteness — NaN/Inf reached the pair bubble and could surface as a saved zero eigenvalue. The loader rejects non-finite content per frequency at the file boundary now (mirroring the susceptibility loader), which also closes the complex-equal_nan discrimination question at its root.

Dead code removed; norb=1 rejection pins and NaN/Inf rejection tests added (the NaN-identical acceptance test is inverted accordingly). Full suite: 1106 passed.

k-yoshimi:
Round 8 caught that the round-7 reorder fixed the helper but not the CALLER: the 2-index route's own norb == 1 branch skipped the helper call entirely, so one-orbital Exchange/PairHop still passed there (reproduced by the review). Fixed in e432c77 — the guard call now precedes that branch for every 2-index input, with route-level tests through `_compute_vertices` (the helper-only test could not see the bypass), allow_ir coverage for the non-finite cases, and the stale comment corrected. Full suite: 1107 passed.

k-yoshimi:
Round 9 (deep) found two real issues, both fixed in 28a469a:

1. **The Exchange/PairHop rejection was an ordering artifact, not an invariant.** It ran only in the orchestration layer, so calling the IR instantaneous-vertex helper directly silently accepted both terms and returned a zero vertex (reproduced at norb=1 and 2). The cheap rejection scan is now split into its own helper and enforced inside `_compute_vertices_flex` itself, so every route -- static, dynamic per-frequency, IR instantaneous -- is guarded by construction. The general (myo) convention is untouched and still accepts both terms.

2. **The frequency-independent S/C matrices were rebuilt by every per-frequency dynamic contraction** (~Nmat identical full-grid pairs per run, ~1000 in production). A single dispatcher now builds one pair per run; the dynamic solver passes it to every contraction and to the IR instantaneous vertex, and the static route shares its pair between the diagnostic and the vertex.

Also renamed stale references to a helper that no longer exists under that name. New tests: boundary-level and route-level rejection (norb=1 and 2), myo acceptance, a one-build-per-run count pin, and an end-to-end static `calc_eliashberg` rejection with Exchange configured. Full suite: 1113 passed.

k-yoshimi:
Round 10 caught a regression introduced by the round-9 caching: supplying precomputed S/C matrices skipped the dispatcher, which was the only place the convention tag was validated -- so an unknown convention with a cache was silently treated as unrestricted and returned a vertex (reproduced). Fixed in eec802c: the tag is normalized and validated before both the rejection and the cache branch, so unknown conventions fail closed at every boundary, cached or not. Also moved the cheap rejection ahead of the large S/C allocation at the three orchestration sites, and added tests: cached-matrices rejection for both terms, and unknown-convention fail-closed at all three boundaries with and without a cache. Full suite: 1115 passed.

PR #122, opened 2026-07-30
Derive the RPA ring's base spin tables from the adjudicated vertex table

k-yoshimi:
## Summary

Phase-2 increment of #107 (after #118 single-sourced the S/C content and the Fierz cross-slot correction): the ring's hand-coded per-type spin tables in \`rpa.py::_make_ham_inter\` -- the density-slot weights and the transverse spin-flip weights adjudicated in #105 -- were the last vertex content living outside \`hwave.solver.vertex_table\`.

- The **density-slot part** is now derived in code as the channel decomposition \`W_same = (C-S)/2\`, \`W_cross = (C+S)/2\` of the adjudicated (S, C) entries, at the slot family each type's base placement occupies: \`diag\` for CoulombIntra, \`antidiag\` for PairHop, \`density\` for the rest.
- The **spin-flip part** (Exchange, PairLift) cannot be derived from the longitudinal channels -- those see only spin-conserving bilinears -- so it is recorded as read-only data (\`RING_SPIN_FLIP\`) next to the table it belongs with.

Any future change to the vertex content now lands in one module for every consumer: the pair-space S/C builders (FLEX general, Eliashberg), the ring's Fierz correction, and the ring's base tensor.

## Zero behavior change

\`ring_spin_table\` reproduces the seven previously hand-coded tables exactly; the new test pins them literal-for-literal (those literals ARE the adjudicated base content) and checks the derived floats multiply bit-identically to the old integer weights. The existing behavior nets -- the one-shot RPA==FLEX equivalence matrix, the transverse-channel suite, and the SCF fixed-point matrix -- pass unchanged. Full suite: 1119 passed, 24 skipped.

Refs #107, #105.


k-yoshimi:
A second, deeper review round confirmed the physics independently: the channel decomposition is the unique algebraic bridge (charge and longitudinal-spin eigenvalues of the spin-conserving block are C and -S under the solver's [1 + chi0*ham]^-1 structure), PairHop's antidiag placement is not a coincidence, and old-vs-new assembly is bit-identical including the aggregate Coulomb, sublattice-folded, and spin-orbital routes. It also flagged one pre-existing item, fixed in e2d4e80: tests/verify_sc_matrices.py was a development-time standalone verifier predating the exact-diagonalization adjudication -- it carried stale copies of the S/C content and the ring spin tables, reported large deviations against the current adjudicated values, and still exited 0. It is removed (nothing referenced it; the adjudicated content is pinned by real tests now), and a synthetic non-integral (S, C) test now demonstrates the decomposition is algebraic rather than a seven-case accident. Full suite: 1120 passed.

PR #123, opened 2026-07-30
Single-source the FFT-grid reversal map across the five solvers

k-yoshimi:
## Summary

First increment of #108 (extract computation shared by FLEX/Eliashberg -- and, it turned out, RPA and UHFk too -- into single-implementation modules).

The momentum/position negation \`i -> (N - i) % N\` on FFT-ordered axes was spelled **three different ways at eighteen call sites**:

- \`roll(arr[::-1], +1)\` -- sc.py (pair bubble, gap parity, k-space symmetrisation via a fancy-index gather), eliashberg_dynamic.py (dynamic pair bubble, gap reversal)
- \`flip(roll(arr, -1))\` -- rpa.py (declaration symmetrisation, longitudinal and transverse chi0 kernels), flex.py (2 IR chi0 kernels), uhfk.py (8 declaration-reversal sites)

kept consistent only by comments ("same reverse+roll map as the uniform path"). The map has a documented trap: for axes of length <= 2 it is the identity, so a misapplied reversal is invisible on one- and two-site fixtures -- exactly the transverse-channel defect history.

\`hwave.solver.kgrid.reverse_fft_axes\` is now the single implementation (numpy and GPU arrays alike). Frequency/imaginary-time reversal deliberately stays local to each caller: its convention is per-grid physics (centered fermionic Matsubara flip, signed tau flip, symmetric IR node flip), and the module documents that boundary.

## Zero behavior change

All three spellings are the same element gather, so consolidation is **bit-identical by construction**. Tests pin: the index formula on odd/even axes, involution, the Gamma fixed point, all three historical spellings bit-for-bit, the short-axis identity trap as executable knowledge, and the two composite site shapes (centered-Matsubara and signed-tau) against the exact old inline expressions. Full suite: 1125 passed, 24 skipped -- including the UHFk reference comparisons and the 13-digit Eliashberg milestone tables, unchanged.

Refs #108.

### DIFF
--- a/src/hwave/sc.py
+++ b/src/hwave/sc.py
@@ -19,6 +19,7 @@ import logging
 import numpy as np
 
 from hwave.solver.vertex_table import sc_coefficients
+from hwave.solver.kgrid import reverse_fft_axes
 from numpy.fft import fftn, ifftn
 from scipy.optimize import bisect
 from scipy.sparse.linalg import LinearOperator, eigs, bicgstab, gmres, lgmres
@@ -839,9 +840,7 @@ def _symmetrise_interactions_k(inter_k):
         if itype == "PairHop":
             out[itype] = 0.5 * (M + np.conj(M.transpose(1, 0, 2, 3, 4)))
             continue
-        nx, ny, nz = M.shape[2], M.shape[3], M.shape[4]
-        Mrev = M[:, :, (-np.arange(nx)) % nx][:, :, :, (-np.arange(ny)) % ny][
-            :, :, :, :, (-np.arange(nz)) % nz]
+        Mrev = reverse_fft_axes(M, (2, 3, 4))
         out[itype] = 0.5 * (M + Mrev.transpose(1, 0, 2, 3, 4))
     return out
 
@@ -2457,11 +2456,9 @@ def _calc_g2(green_kw, beta):
     Nx, Ny, Nz, nmat = green_kw.shape[2], green_kw.shape[3], green_kw.shape[4], green_kw.shape[5]
     nvol = Nx * Ny * Nz
 
-    # G(-k, -wn) via roll+flip
-    green_kw_inv = np.roll(
-        green_kw[:, :, ::-1, ::-1, ::-1, ::-1],
-        (1, 1, 1), (2, 3, 4)
-    )
+    # G(-k, -wn): centered-Matsubara flip on the frequency axis, then the
+    # shared FFT-grid map k -> -k on the spatial axes.
+    green_kw_inv = reverse_fft_axes(green_kw[..., ::-1], (2, 3, 4))
     # einsum("ijpqsk, lmpqsk -> ijlmpqs") sums over k (nmat dimension)
     # Reshape to use tensordot for BLAS: contract last axis (nmat) after
     # merging spatial dims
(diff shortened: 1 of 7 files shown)


k-yoshimi:
The first review round verified all eighteen conversions byte-for-byte on their actual layouts (including the tau-axis handling in both RPA kernels, the commutation of the FLEX tau flip with the spatial map, and all eight UHF transpose/conjugate chains) and found no correctness issue. It flagged two documentation inaccuracies, fixed in 36294fa: the module docstring undercounted the call sites, and its axis-scope paragraph contradicted the RPA kernels' legitimate use of the map on the uniform imaginary-time grid -- the helper is now described as a generic periodic-grid index reversal with the non-applicable frequency conventions listed explicitly. Also added production-facing pins that call the converted gap-reversal and symmetrisation functions directly against the exact pre-consolidation expressions. Full suite: 1128 passed.

k-yoshimi:
Round 2 found a nineteenth spelling of the reversal that both my sweep and the first round missed: the transverse vertex assembly computed q -> -q through a coordinate-wise flat-index formula. Consolidated in e5fd761 -- unflatten, apply the shared map, flatten back; the same permutation, so bit-identical -- with a production-facing pin of the assembly method on a 4x3x5 lattice against a replica of the exact old gather (every dimension > 2, so the short-axis identity cannot mask a defect). The module docstring count is corrected and an unused import removed. Full suite: 1129 passed.

k-yoshimi:
Third review round: functionally converged, no correctness regression; independent verification confirmed the transverse consolidation shares memory with the old gather result exactly, the production pin fails under both attempted mutations (swapped unflatten order; flip without roll), and a tree-wide sweep found no further spelling of the map. Fixed the one remaining textual item (the test docstring's stale count) and added the recommended CUDA-conditional device-array test, which skips without a usable device. Full suite: 1129 passed, 1 conditionally skipped.

k-yoshimi:
The previous commit broke CI on all four version jobs: the new device-conditional test used a pytest-style skip inside a unittest.TestCase, and CI runs the suite with unittest, where that skip mechanism registers as an ERROR (the existing GPU tests never hit this because their module-level function style is invisible to unittest discovery). Fixed in c938b4c with unittest-native skips, verified under both runners locally.

k-yoshimi:
A fourth, deeper review round re-derived the reversal from the operator definitions at two physically distinct sites (the longitudinal chi0 kernel and the UHFk declaration partner) -- an equality-only proof would also have preserved a pre-existing bug, and none was found. Two items fixed in ba87eae: an empty axes tuple silently no-opped where the caller believed a reversal happened (now ValueError, with a test), and seven production comments still described the removed spellings in the present tense (now reference the shared reversal; historical defect notes keep their wording). The round also corrected the record on the CI incident: the workflow runs unittest as the gating step and then pytest non-gating for coverage, so pytest-style module-level tests do execute but cannot gate. Full suite: 1130 passed, verified under both runners.

k-yoshimi:
Round 5 verified the empty-axes rejection (all nineteen production call sites pass non-empty literal tuples), confirmed the comment edits changed no executable code (AST equality against the prior commit), and found exactly one remaining present-tense old-spelling docstring, now fixed -- a docstring-only change. RPA ladder and FLEX general suites re-verified in the round (32/32 and 11/11).

PR #124, opened 2026-07-31
Pin the transverse channel's reduced-chi0q guards with direct tests

### LINKED ISSUE #111: Unreachable reduced-chi0q expansion branches in _build_transverse_channel
`RPA._build_transverse_channel` contains expansion branches for a REDUCED chi0q (4-dim spin-free and 4-dim spin-diag), but `_set_scheme` exits with an error unless `calc_scheme` is `general` whenever `calc_type = "ring+ladder"`, so these branches are unreachable through any normally constructed solver. They can only be entered by direct private calls or hand-built stubs (one test does exactly that for the error path).

Dead defensive code in a channel that has just been through six review rounds is a liability: it looks supported, it is untested, and it re-implements the reduced-to-general expansion found elsewhere.

Options, from the #105 review: mark the branches explicitly as private/defensive and add direct tests, or remove them in a small compatibility-reviewed change.

k-yoshimi:
## Summary

Closes #111. The dead reduced-expansion branches in \`_build_transverse_channel\` -- both carrying the same defective \`delta_{l2,l4}\` placement the Eliashberg loader had, both unreachable because \`ring+ladder\` forces \`calc_scheme='general'\` -- were already removed on the density-pair embedding branch and merged with PR #121, replaced by loud guards.

What remained of the issue's ask was direct test coverage of those guards, since no normally constructed solver can reach them. This adds two tests using the same private-stub pattern the existing spin-diag guard test uses:

- the spin-free branch rejects a reduced (2-index) chi0q, and its message names the correct density-pair embedding a future relaxation would need;
- the spinful branch rejects a non-6-dim chi0q, which cannot supply the spin-flip block.

Test-only change; verified under both the gating unittest runner and pytest. Full suite: 1132 passed.


k-yoshimi:
First review round: no correctness issue; two test-quality items fixed in e28928c -- the intentional exception-type asymmetry between the two guards is now documented in the test docstring (AssertionError for the invariant guaranteed by normal construction, ValueError for the shape that depends on runtime input), and the spinful assertion now pins the intended failure (channel, requirement, scheme, shape) rather than any ValueError containing 'full'. Also dropped an unread stub attribute flagged by the round.

k-yoshimi:
A third, deeper review round independently re-derived the dead-branch claim from the code (single production caller, the construction-time scheme gate, and shape validation of external bubbles -- no supported configuration could reach the removed branches, so no compatibility note is needed) and found that the exception-type asymmetry my earlier commit had documented as intentional rested on a false distinction. Fixed in 4a3124f: the spin-free guard now raises ValueError consistently with the other two guards, the spinful stub models the state a constructed solver could actually be in (ns=2), and a new table-driven test pins the upstream invariant itself -- reduced/squashed exit at construction, auto resolves to general, and a spy confirms every normally reached transverse call carries the full 6-dim bubble. Full suite: 1133 passed, both runners.

k-yoshimi:
Round 4 caught that the new gate test documented a false universal: it asserted every normal transverse call carries a 6-dim bubble, but a valid spin-diagonal ring+ladder run reaches the channel with a 7-dim bubble (leading block axis) -- the test only exercised the spin-free default. Fixed in 15bd0e3: the spy is table-driven over all three spin modes with their measured ranks (spin-free 6, spin-diag 7, spinful 6), using programmatic spin-orbital fixtures for the latter two. The round also confirmed the guard-exception harmonization breaks no caller or test. Full suite: 1133 passed, both runners.

k-yoshimi:
Round 5 verified the per-mode fixtures produce their intended modes and ranks by direct execution, and caught two remnants of the round-4 correction, fixed in 1e160cd: the docstring's opening sentence still carried the false universal '6-dim' claim, and the spy recorded rank alone -- a spinful fixture silently regressing to spin-free would have passed. The spy now records (spin mode, rank) pairs and each case asserts its exact pair.

k-yoshimi:
A seventh, deeper round found one real production defect the previous six missed: the transverse bubble kernel guarded its two-spin-block requirement with a bare assert, which disappears under python -O -- a three-block Green's function was then accepted with the extra block silently ignored. Fixed in 456c551 as a ValueError naming the observed count, with a regression test that also runs under -O. The round's test-hardening items are in too: the spinful case asserts the documented Sz-conserving-block warning (shape/reachability, not physics), the construction-gate checks assert exit code and message rather than any SystemExit, both spin-orbital solves must yield entirely finite transverse output, and the clean spin-diagonal case pins a compact physical property (static on-site response real and positive). Full suite: 1134 passed, both runners.

k-yoshimi:
Two independent review passes ran in parallel for this round. The first measured the new physical pins' margins (imaginary residue five orders below tolerance, positivity nine orders above zero, bit-identical across BLAS thread counts) and confirmed no caller can legitimately deliver a wrong block count nor catch the new error into different behavior. The second ran mutation testing: reverting each of the three production changes is killed by a committed test, and every new test kills its own mutation. Fixed in 68c628d from their findings: a permanent optimized-Python subprocess regression for the block-count guard (the gating CI never runs -O, so an assert regression would otherwise pass CI), an accuracy fix to the guard-harmonization comment (the spin-diag checks reject reachable provenance conditions, not malformed ranks), and the EN/JA algorithm docs now state the Sz-conserving-block limitation for spin-mixing transverse runs, matching the runtime warning. Full suite: 1135 passed, both runners.

PR #126, opened 2026-07-31
Make pytest read pytest.ini, and make the CI coverage gate actually gate

### LINKED ISSUE #84: pytest.ini uses a [tool:pytest] header, so its settings (incl. the coverage gate) are silently ignored
## Summary

`pytest.ini` uses a `[tool:pytest]` section header. That is the header pytest reads from **`setup.cfg`** — in a standalone `pytest.ini` the correct header is `[pytest]`. As a result pytest applies **none** of the file's settings: `addopts`, `markers`, and everything else are silently ignored.

The practical consequence is that the coverage gate the CI appears to enforce is **not actually enforced**: `--cov-fail-under=80` lives in `addopts` and never reaches pytest. A change that drops coverage below the threshold will pass CI.

## How it was noticed

While adding a `slow` marker for an on-demand fixture-regeneration test (#82), registering the marker in `pytest.ini` had no effect — the marker stayed unrecognised. Adding a minimal `tests/conftest.py` that registers it programmatically worked, which is the workaround currently in place.

## Suggested fix

Rename the header to `[pytest]` (or move the contents into `setup.cfg`/`pyproject.toml` under the appropriate key).

**Please land this deliberately rather than as a drive-by**: it will switch on every option in the file at once, including the coverage gate. It is worth checking first what the real coverage number is, and whether the other `addopts` entries are still wanted, so the flip does not turn CI red unexpectedly.

Found while working on #82; not caused by it.

k-yoshimi:
## Summary

Closes #84. \`pytest.ini\` used the \`[tool:pytest]\` header -- the one pytest reads only from \`setup.cfg\` -- so in a standalone \`pytest.ini\` **every setting in the file was silently ignored**: \`addopts\`, the marker registry, and the \`--cov-fail-under=80\` floor the CI appeared to enforce.

Per the issue's request this lands deliberately, not as a drive-by header flip:

- **Header corrected** to \`[pytest]\`; the file documents its own failure mode and its two constraints.
- **addopts slimmed** to \`--tb=short --strict-markers\`. The coverage flags are removed from the ini: \`run_tests.yml\` runs bare \`pytest\` without \`pytest-cov\` installed, and activating \`--cov\` from the ini would break that workflow. \`-v\` (noise at 1100+ tests) and \`--disable-warnings\` (hides real signals) are dropped.
- **Markers genuinely register** now, with \`--strict-markers\` making any unregistered marker fail loudly.
- **The coverage floor moves to where coverage actually runs**: the \`ci-python39.yml\` pytest step gains \`--cov-fail-under=80\` and loses \`continue-on-error\`, so it now gates. This also closes a known gap: module-level pytest-style tests are invisible to unittest discovery and previously ran only in a step whose failures could not fail CI.

## Verification

Measured before the flip: coverage 84.6% (floor 80, margin ~4.6 points). Full suite green under the new options with both runners: pytest 1135 passed / 25 skipped, unittest discovery OK. The exact gated CI command (cov + fail-under) passes locally.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,9 @@
+# Report precision for the coverage gate (PR #126, round-3 review).
+# coverage compares --cov-fail-under against the ROUNDED total: at the
+# default zero-decimal precision, 79.50% rounds to 80 and PASSES an
+# 80-percent floor. Two decimals shrink that rounding leeway to 0.005
+# points. Kept in .coveragerc (not pyproject.toml) deliberately: reading
+# coverage config from pyproject requires the coverage[toml] extra on
+# Python < 3.11, and a missing extra would silently drop this setting.
+[report]
+precision = 2
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,9 +1,10 @@
-# Report precision for the coverage gate (PR #126, round-3 review).
-# coverage compares --cov-fail-under against the ROUNDED total: at the
-# default zero-decimal precision, 79.50% rounds to 80 and PASSES an
-# 80-percent floor. Two decimals shrink that rounding leeway to 0.005
-# points. Kept in .coveragerc (not pyproject.toml) deliberately: reading
-# coverage config from pyproject requires the coverage[toml] extra on
-# Python < 3.11, and a missing extra would silently drop this setting.
+# Report precision (PR #126). Two decimals make the DISPLAYED totals and
+# the --cov-fail-under comparison finer-grained, but any finite precision
+# still rounds: the HARD floor is enforced separately by
+# tests/_coverage_floor.py, which compares the integer line counts from
+# coverage.xml exactly. Kept in .coveragerc (ini format, read
+# unconditionally) rather than pyproject.toml so the setting does not
+# depend on how coverage's TOML support is provided in a given
+# environment.
 [report]
 precision = 2
--- a/.github/workflows/ci-python39.yml
+++ b/.github/workflows/ci-python39.yml
@@ -67,6 +67,7 @@ jobs:
       # tests nor the gate could actually fail CI).
       run: |
         poetry run pytest tests/ -v --cov=src/hwave --cov-report=term-missing --cov-report=xml --cov-fail-under=80
+        poetry run python tests/_coverage_floor.py coverage.xml 80
 
     - name: Verify the test suite left the tracked tree clean
       # !cancelled(): report isolation defects even when the gating test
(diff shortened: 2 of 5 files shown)


k-yoshimi:
First review round converged with no findings; its verification covered the activation blast radius (collected-test counts identical with and without the ini, all pytest entry points enumerated and unaffected, the only custom marker registered, GPU tests skip cleanly without a device, coverage path resolution unchanged, and the step-failure semantics downstream of the removed continue-on-error). Its one committed suggestion is added in 2fedcf4: a regression test that asserts the ini was actually APPLIED (resolved path, marker registry, addopts) rather than merely that collection worked -- mutation-verified to fail if the header ever regresses. The full version matrix including the new gating coverage step is green.

k-yoshimi:
A third, deeper round found a real hole in the floor: coverage compares the threshold against the ROUNDED total, so at the default zero-decimal precision 79.50% rounds to 80 and passes -- measured against coverage's own gate function. Fixed in 926d903: .coveragerc sets two-decimal precision (deliberately not pyproject.toml -- coverage reads that only with the toml extra on Python < 3.11, and a missing extra would drop the setting silently), with a boundary test pinning 79.99-fails/80.00-passes, the default-precision leak, and that the config actually loads. The round also verified the checkout-layout safety of the bare-pytest workflow and the rootdir resolution; its two optional items are in (full-path ini assertion; the tracked-tree cleanliness step now runs even when the gating step fails, when its diagnostic matters most). Full suite: 1137 passed, gate at 84.60%.

k-yoshimi:
A fifth, deeper round found the floor still leaked: any finite report precision rounds, so coverage in [79.995, 80) passed -- and 6151/7689 = 79.9974% is only three added statements away in this tree, making the hole attainable rather than theoretical. It also found the boundary test bound to undocumented coverage internals, which an upgrade could rename (silently skipping the test under the ImportError guard). Fixed in b73feb8: the hard gate is now an exact integer comparison over the line counts in coverage.xml (pure stdlib, no rounding), run by CI right after the pytest step and verified to FAIL on the review's own attack value while --cov-fail-under would have rounded it to a pass; the tests pin the exact predicate and use only coverage's public get_option API; two comment-accuracy items are corrected. Full suite: 1138 passed, exact floor at 84.5954%.

k-yoshimi:
Sixth round converged with no findings; its verification covered the checker's malformed- and missing-XML behavior (both loud, exit 1) and the runner semantics. Its one non-blocking suggestion is added: CLI-level subprocess tests pinning the checker's exit codes for passing, failing (including the rounding-attack value), malformed and missing XML.

PR #127, opened 2026-07-31
Longitudinal chi0q shape guards survive optimized Python

### LINKED ISSUE #125: Longitudinal chi0 kernel validates array shapes with bare asserts, which vanish under python -O
Split out of the PR #124 review, as the generalization of the defect fixed there.

`RPA._calc_chi0q` (around src/hwave/solver/rpa.py:2496) validates its input array dimensions (`nvol`/`nmat` consistency) with bare `assert` statements. Like the transverse block-count guard fixed in PR #124, these are DATA-boundary checks, not internal postconditions: under `python -O` they disappear, and a malformed Green's function or tail array would proceed into the kernel instead of failing loudly -- the class of silent wrong-output defect this project has been eliminating.

Fix: replace them with `ValueError`s reporting the expected and observed values, with direct regression tests that also run under `-O` (see the pattern in PR #124's `test_transverse_bubble_rejects_wrong_block_count`).

Note from the same review: the assertions at rpa.py lines ~1307, ~1310 and ~2415 are internal postconditions rather than independent data boundaries, and can stay as asserts.

k-yoshimi:
## Summary

Closes #125, the longitudinal analogue of the transverse block-count fix from PR #124: \`RPA._calc_chi0q\` validated its Green's function's volume and frequency axes with bare \`assert\`s, which vanish under \`python -O\` -- a malformed array would then proceed into the kernel and produce plausible-looking wrong output instead of failing loudly.

Both guards are now \`ValueError\`s naming the expected and observed values. Regression tests pin them under the normal interpreter and in a \`-O\` subprocess (the gating CI never runs optimized Python, so without that pin a regression back to \`assert\` would pass CI). Same stub pattern and rationale as the transverse guard tests.

The assertions at rpa.py ~1307/~1310/~2415 identified in the review as internal postconditions (not data boundaries) are intentionally left as asserts.

Full suite: 1142 passed, both runners.

### DIFF
--- a/src/hwave/solver/rpa.py
+++ b/src/hwave/solver/rpa.py
@@ -1302,13 +1302,22 @@ class RPA:
                 chi0q = chi0q[:,self.freq_index]
                 logger.info("filter range in matsubara frequency: {} in {}".format(chi0q.shape[0], self.nmat))
 
-            # nblock
+            # nblock: defense in depth behind the kernel's structural
+            # check -- under python -O the former asserts vanished and a
+            # wrong block count was silently truncated (round-5 review)
             if self.spin_mode in [ "spin-free", "spinful" ]:
-                assert chi0q.shape[0] == 1
+                if chi0q.shape[0] != 1:
+                    raise ValueError(
+                        "chi0q block count {} does not match spin_mode "
+                        "'{}' (expected 1)".format(
+                            chi0q.shape[0], self.spin_mode))
                 chi0q = chi0q[0]
             else:
-                assert chi0q.shape[0] == 2
-                pass
+                if chi0q.shape[0] != 2:
+                    raise ValueError(
+                        "chi0q block count {} does not match spin_mode "
+                        "'{}' (expected 2)".format(
+                            chi0q.shape[0], self.spin_mode))
 
             green_info["chi0q"] = _bk.to_host(chi0q)
             # Record the producing run's frequency metadata alongside: a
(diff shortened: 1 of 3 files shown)


k-yoshimi:
The first review round found no legitimate caller that can trigger the new guards (the kernel always receives the full grid; frequency filtering happens downstream), but it broke my regression test twice, both fixed in a86c44c: the message validation inside the -O subprocess was itself an assert -- optimized away, so a reverted volume guard passed via an unrelated reshape error (reproduced) -- and the message assertions accepted the observed digits anywhere, surviving a mutated expectation. The subprocess now uses explicit if/raise with one guard per subprocess, the assertions bind role-bound fragments, both mutations are verified killed, and the guard comment no longer overstates the volume failure mode. All three CI gates verified locally.

k-yoshimi:
A third, deeper round found the real gap the first two missed: the kernels validated the Green's function but took green0_tail on faith, and a tail of the same element count with a different shape is silently reshaped into finite, plausible-looking chi0q (reproduced with a measured magnitude). Every normal call site passes the pair one _calc_green call returned, so no working configuration changes -- but the kernel accepts two independent arrays, and for research software that invariant should be enforced, not assumed. Fixed in afd1a49: exact shape equality between tail and Green's function in BOTH the longitudinal and transverse kernels, with same-element-count regression tests for each, plus a docstring accuracy fix. The round also confirmed the guards are GPU-safe (shape entries are host integers; the error propagates before any device work). Full suite: 1145 passed, both runners.

k-yoshimi:
Round 4 converged with no findings: the round-3 reproduction now raises in both kernels before any reshape, the guard ordering yields the most informative error for each malformed input, and all legitimate call paths (RPA solve, FLEX uniform including coeff_tail=0, the SC internal route, the transverse pair) pass untouched. Its one non-blocking suggestion -- pinning the error precedence on combined malformations -- is added.

k-yoshimi:
A fifth, deeper round reproduced fail-open structural inputs the first four missed: the longitudinal kernel accepted a zero-block Green's function (finite empty result) and a three-block one -- which the caller's bare block asserts then silently truncated under optimized Python, exactly this PR's failure model -- plus zero-extent and rectangular orbital axes; the transverse kernel processed a wrong frequency count normally. Fixed in 9d24dfe: both kernels validate the block axis, the orbital square, and (transverse) volume/frequency, and the caller's block handling replaces its asserts with spin-mode-keyed errors as defense in depth. The round also caught the earlier transverse -O regression test using an in-subprocess assert (the exact flaw the newer helper documents; now explicit if/raise) and sharpened the tail-guard comment: shape equality is structural, not provenance. Full suite: 1148 passed, both runners.

k-yoshimi:
Round 6 converged with no findings; all four round-5 reproductions were rerun under optimized Python against the fixed kernels and every legitimate spin mode passes end to end. Its non-blocking suggestion for persistent optimized-Python coverage of the new structural guards is added via the existing subprocess helper.

Issue #78, opened 2026-07-23
hwave_sc: general-FLEX chiq_s/chiq_c stored transposed → chi0q_mode='flex' disagrees with 'load'


k-yoshimi:
## Summary

For an identical physical setup (self-energy Σ = 0), the linearized Eliashberg solver `hwave_sc` returns a different static eigenvalue λ depending on how the susceptibility is supplied:

| input mode | static λ |
|---|---|
| `chi0q_mode = "load"` (RPA χ⁰, χ_s/χ_c rebuilt internally) | 0.2269804250 |
| `chi0q_mode = "flex"` (χ_s/χ_c read from a general FLEX run) | 0.2219819770 |

Two-orbital Hubbard (`CoulombIntra` only), 8×8, `Nmat=64`, singlet, static. The FLEX run uses `IterationMax=1, Mix=0.0`, so Σ = 0 and the two inputs describe the *same* physics — they must give the same λ. The `load` value is the correct one.

## Root cause

The general (full-vertex) FLEX path works internally in the MYO orbital-pair convention, which is the orbital-pair transpose of the RPA `[a,c,b,d]` layout used everywhere else in the Eliashberg solver.

At the general-FLEX output boundary (`src/hwave/solver/flex.py`, `_flex_compute_veff_general`), `chi0q` is transposed back to `[a,c,b,d]` for output, but `chi_s`/`chi_c` are written still in MYO layout:

```python
chi0q_out = chi0q.transpose(0, 1, 4, 5, 2, 3)   # chi0q -> [a,c,b,d]
return chi0q_out, v_eff, chi_s, chi_c            # chi_s/chi_c left in MYO
```

The reader (`src/hwave/sc.py`) consumes them without transposing: `_expand_flex_chi` passes the MYO arrays through, and `_compute_vertices_flex` builds the pairing vertex `S·χ·S` assuming the native `[a,c,b,d]` orbital-pair layout. The `chi_convention="myo"` tag only selects the S/C interaction matrices, which differ from the native ones by a single charge element (`-U'+2J` vs `-U'+J`, zero for pure Hubbard) — it does **not** correct the layout.

As a result the pairing vertex built from FLEX input is the orbital-pair transpose of the correct one, and the eigenvalue is wrong. The on-disk FLEX output is also internally inconsistent: `chi0q` is stored in `[a,c,b,d]` while `chiq_s`/`chiq_c` are stored transposed.

## Confirmation

[...]

Issue #85, opened 2026-07-26
RPA chi0q bubble is not run-to-run reproducible (~1e-17), defeating byte-level pinning


k-yoshimi:
## Summary

`RPA._calc_chi0q` / the internal bubble used by `sc._calc_chi0q_internal` is **not run-to-run reproducible**. Two identical invocations on identical input differ at the ~1e-17 level.

Measured (same process, same input, back-to-back):

```
bit-identical: False
max|Δ| real = 1.39e-17
max|Δ| imag = 3.58e-18      (for reference, max|Im| = 1.42e-15)
```

The drift affects both the real and imaginary parts, and it **persists with `OMP_NUM_THREADS=OPENBLAS_NUM_THREADS=MKL_NUM_THREADS=VECLIB_MAXIMUM_THREADS=1`**, so it is not BLAS thread-count nondeterminism — most likely ULP/alignment-dependent SIMD paths inside the FFT/GEMM calls.

## Why it matters

The magnitude is physically irrelevant, but it defeats byte-level pinning:

- regression tests cannot assert bit-identical end-to-end output for anything downstream of `chi0q` unless `chi0q` itself is pinned (`chi0q_mode="load"`);
- `hwave_tsweep` result diffing across runs shows spurious differences;
- CI pinning strategies that assume determinism will be flaky.

This was worked around in #82 by pinning `chi0q` via `chi0q_mode="load"` for the one regression test that needed byte-identity; everything downstream of a fixed `chi0q` **is** bit-reproducible (verified).

## Suggested next step

Localise which call in the bubble is responsible (FFT vs the batched GEMM) and decide whether it is worth forcing a deterministic path there, or whether the project prefers to document that only tolerance-based comparisons are meaningful for `chi0q` and anything derived from it.

Found while working on #82; pre-existing, not caused by that PR (the bond path never calls this function).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

k-yoshimi:
Closing: the non-reproducibility no longer exists on the current unified bubble kernel. Re-measured today on develop (bbb8637f, macOS/Accelerate — the same class of machine as the original measurement):

- kernel level (`bubble.dense_bubble`), synthetic inputs, 64×32 and 128×64-cell/nd=4 (~34 MB output): back-to-back runs **bit-identical**, three repetitions;
- `bubble.contract_general` and `backend.spatial_ifftn` stages individually: bit-identical;
[...]

PR #128, opened 2026-07-31
Reject legacy general-path chi files when the interaction orientation matters

### LINKED ISSUE #101: Pre-#99 FLEX susceptibility files record no interaction orientation, so hwave_sc cannot detect a mismatch
Split out of the #96 review (PR #99).

## The gap

PR #99 changes the orbital orientation in which `sc._build_interaction_k` stores the interaction. That builder is used by three consumers, one of which is the **general FLEX path** (`solver/flex.py`), which builds its MYO S/C matrices from it and then writes `chiq_s`/`chiq_c`.

So a susceptibility file produced by a pre-#99 general-FLEX run was computed with the old orientation, while `hwave_sc` reading it now assembles the pairing vertex with S/C matrices built in the new one. Nothing detects the mismatch: `chi_convention = "myo"` records which S/C *convention* the file pairs with, and `chi_orbital_layout = "acbd"` records the on-disk index order, but neither records the interaction orientation.

Affected readers: `_read_flex_chi_raw` / `_compute_vertices_flex` (static), and `solver/eliashberg_dynamic.py` for both the uniform and IR dynamic paths, including the separately constructed instantaneous vertex.

## Scope

Narrow but real. The orientation only matters when the interaction is not symmetric under `a <-> b`. The general FLEX path rejects off-site two-body terms, so the reachable case is an **asymmetric on-site** inter-orbital entry — which the readers accept without validation (#93), and which does reach the S/C builders (verified in `tests/test_vertex_orientation.py::TestAllInteractionTypesTransposed::test_asymmetric_on_site_input_reaches_the_sc_builders`).

For a symmetric interaction — every fixture in the tree, and any physically ordinary input — old and new files are interchangeable.

## Options

1. Stamp an interaction-orientation (or package-version) marker into the FLEX susceptibility outputs and validate it in the shared loader, rejecting a legacy file when the configured interaction is not invariant under the orbital transpose. This is the same shape as the `chi_orbital_layout` guard added for #78.
2. Rely on #93: once interaction files are required to be orbital-symmetric, the ambiguity disappears and no marker is needed.

Option 2 is cheaper and removes the class of problem rather than detecting it, so it may be worth settling #93 first.

Tests to add either way: legacy MYO fixtures without the marker, checked for rejection with an asymmetric interaction and for acceptance with a transpose-invariant one, across the static, uniform-dynamic and IR-dynamic paths.

Related: #96, #93, #78.

k-yoshimi:
## Summary

Closes #101. PR #99 changed the orbital orientation \`sc._build_interaction_k\` stores -- the orientation the general (myo) S/C matrices consume. A myo susceptibility file that carries the #78 \`chi_orbital_layout\` marker but no \`sc_vertex_version\` was produced in the window between those two changes and **passes every existing gate while carrying the old orientation**; \`hwave_sc\` would silently assemble the pairing vertex in the new one (git-dated: the layout marker landed 2026-07-23, the orientation change 2026-07-28, the version tag 2026-07-30).

Rather than introducing a new marker, this uses a chronological fact: **version-2 files are necessarily post-#99**, so for a myo file the existing version requirement doubles as the orientation marker. The no-version rejection now also fires when the configured interactions declare an asymmetric on-site inter-orbital coupling -- the one reachable orientation-sensitive case (the general path rejects off-site two-body terms; PairHop is excluded because its declaration partner is the Hermitian entry; reduced/kuroki files never depended on this builder's orientation and stay accepted). The message names each applicable reason (#113 content, #101 orientation, or both).

The check lives in the shared loader, so the static, uniform-dynamic and IR readers are all covered, matching the issue's reader list.

## Tests

Rejection with an asymmetric on-site coupling through both the raw shared loader and the static loader route; acceptance for a symmetric myo file, a legacy kuroki file with the same asymmetric coupling (behavior unchanged), and a version-2 myo file with the asymmetric coupling; and the combined-reasons message. Full suite: 1152 passed, both runners.


k-yoshimi:
The first review round confirmed the chronological premise against git history (no version-stamping code predates the orientation change on any branch) and the mutation coverage, and found three real items, fixed in e558ad7: PairLift was in the orientation predicate although its particle-hole vertex contribution is exactly zero on both sides -- rejecting a compatible file on an inert term; the rejection message overclaimed that every unversioned file is pre-#99 (a two-day window of files is correctly oriented but predates the declaration symmetrisation that arrived with the version stamp -- the message now states the semantics cannot be verified, which is the true reason and covers both); and the asymmetry helper failed OPEN on NaN couplings because max() drops NaN. The EN/JA docs now describe the extended rule, and the round's recommended tests are all in, including a parser-backed key-type case and the allow_ir routes. Full suite: 1159 passed.

k-yoshimi:
Round 2 verified all code-side fixes by direct execution (including an independent mutation of the predicate, killed by the helper regression) and found the user-facing docs contradicting the implementation: they claimed ANY configured interaction triggers the rejection, while the predicate names four types and the tests require PairLift/PairHop acceptance. Fixed in 23a1da5 -- the docs name the four types, explain both exclusions, and use the historically neutral 'cannot be verified' framing; the message regression now pins that wording, and the helper gains +inf and complex-NaN fail-closed cases.

k-yoshimi:
Round 3 converged: the documentation now matches the implementation exactly in both languages (types, exclusions and their reasons, neutral timing framing), verified alongside a rerun mutation. Its non-blocking suggestion is added -- a membership pin asserting the orientation-specific wording for each of the four types, since Hund/Ising/Exchange would otherwise be rejected by the older content guard and a dropped list entry could escape detection.

k-yoshimi:
A fourth, deeper round measured its way to two real defects the first three missed. First, PairHop's wholesale exemption was wrong: its orientation indeed never changed, but the conjugated-mean reading of its Hermitian-partner declarations arrived with the version stamp, and a non-Hermitian-closed on-site declaration changes the general-scheme S/C matrices by a measured 0.15 -- the acceptance test was pinning unsafe behavior. Unversioned files with such declarations are now rejected under a declaration-semantics reason; Hermitian-closed pairs and version-2 files stay accepted, and the docs state the corrected rule. Second, the standalone reader matched interaction keys with exact case while the FLEX reader is case-insensitive, so a lowercase-configured run produced susceptibilities with an interaction the Eliashberg side silently read as absent -- bypassing this gate and building the vertex from nothing; the reader is case-insensitive now. The round's remaining item -- sublattice folding being accidentally unsupported and failing late -- becomes an early actionable rejection at both entries. All fixed in 7f364cf; full suite: 1163 passed.

k-yoshimi:
Round 5 found the case-insensitivity fix incomplete: the chi0q_mode='calc' auto selector still checked the raw configuration with exact case, so a lowercase-configured run silently chose the reduced scheme and omitted the components its own logic says inter-orbital interactions require. Fixed in dfeb7a6 -- the selector lookup is case-insensitive and the aggregate Coulomb keyword now counts as inter-orbital (its split can produce an inter-orbital part). The round's recommended tests are in: public-entry parity between canonical and lowercase keys on saved eigenvalues (numeric tolerance -- the byte-identity attempt surfaced the known ~1e-17 eigensolver non-reproducibility, issue #85), the measured 0.15 PairHop basis pinned against the general-scheme matrices, and a sentinel proving the SubShape rejection precedes any file access. Full suite: 1166 passed.

k-yoshimi:
Round 6's systematic sweep surfaced the case-defect class twice more, fixed in e0481f7. The temperature-sweep resume fingerprint performed case-sensitive lookups on the interaction table: a lowercase key executed correctly but was left out of the filename and content-digest fields, so a resume could silently reuse results computed with a different interaction file -- reproduced by swapping the file behind an unchanged fingerprint. And the k-space input reader validated keys case-insensitively but dispatched on exact case, feeding a lowercase-configured geometry file to the interaction parser. Both lookups are normalized, with gating unittest coverage: every fingerprint interaction key pinned lowercase-participating (canonical equals lowercase, and an in-place edit invalidates either), and reader parity for lowercase Geometry/Transfer. Full suite: 1168 passed.

k-yoshimi:
Round 7's repeated sweep surfaced the case-defect class a fifth time -- the sublattice reshape dispatched on exact-case names, so a lowercase-configured geometry table crashed in the interaction reshaper -- and found the SubShape guard keyed on the explicit key while the package convention defaults an omitted SubShape to CellShape, a fully folded configuration; the omitted case reached the file reader with exactly the mismatch the guard exists to stop. Fixed in f68a998: normalized dispatch in the reshape and the reader's path lookup, and the guard now acts on the RESOLVED value with the convention explained in its message (an intermediate attempt to change the default to [1,1,1] was itself wrong -- it contradicted the documented package convention -- and was corrected before commit). Gating tests: sublattice-reshape parity for lowercase keys, mixed-case path resolution, and the omitted-SubShape rejection. Full suite: 1171 passed.

k-yoshimi:
Round 8's final sweep passed 112 canonical/lower/mixed-case lookup comparisons across the reader, Eliashberg loading, auto-scheme selection, fingerprinting, and folded interactions -- and surfaced the class a sixth time on its way: run-time path resolution in the temperature sweep absolutized only an exact-case key, so a mixed-case configuration hashed the right file while executing against the unresolved path. Fixed in 1bf6773 with a regression that stops the run right after the resolution block and asserts the resolver saw the mixed-case relative path. Round 7's reproductions were re-verified against the fixed code (folded parity for lowercase keys; the omitted-SubShape sentinel with zero reader calls). Full suite: 1172 passed.

k-yoshimi:
Round 9 red/green-tested the round-8 regression and proved it a false positive: the fingerprint makes the identical resolver call before the run-time block, so the spy test passed against the broken parent commit. Replaced in b6cfde4 with an execution-bound capture -- the dict handed to the per-rung solver must carry the mixed-case path absolutized -- verified to fail on the parent and pass on the fix, keeping the non-mutation check. The docs contradiction it flagged (symmetric-declarations wording vs the PairHop Hermitian-closed rule) is fixed per rule in both languages. The round's full case sweep came back clean across reader, selector, fingerprint, dispatch, reshape and run-time paths, and the whole-diff coherence pass found no contradictions. Full suite: 1172 passed.

k-yoshimi:
Round 10 converged: independent red/green reproduction of the execution-bound path regression, a seven-case legacy compatibility matrix, and the coherence pass over the full ten-round diff all passed with no findings. Its two cosmetic notes (a stale test docstring; guard formatting) are applied.

PR #129, opened 2026-08-01
Single-source the declaration symmetrisation in both of its representations

k-yoshimi:
## Summary

Increment 2 of #108. The reduction of double-ended interaction declarations to their physical symmetric coefficient -- the mean with the reversed-bond partner \`(R, a, b) <-> (-R, b, a)\`, and PairHop's Hermitian-partner variant -- existed as two implementations kept aligned only by comments: real-space dense in the RPA ring assembly, momentum-space in the pair-space S/C builders (RPA/FLEX/Eliashberg vertex side). This duplication class is exactly where #96/#106/#112 defects came from.

\`hwave.solver.declarations\` now owns the content in **both representations**, with the design decision documented in the module: each consumer keeps its historical bit-exact pipeline (unifying to literally one representation would change float summation order in one consumer), and the identity that makes them one reduction --

    fft(symmetrise_dense(T)) == symmetrise_k(fft(T))

-- is pinned **executably** for both rules on odd/even lattice dimensions, instead of living in prose. This mirrors the pattern of the adjudicated vertex table (#118): one module owns the physics; derivations are code.

## Zero behavior change

Both delegations are bit-identical: the ring assembly passes its dense array to the shared core (same operations, same order), and the k-space body moved verbatim. Tests additionally pin bit-parity of each form against the pre-refactor inline expressions, the one-sided-bond split (half weight at the wrapped \`-x\` cell on the transposed pair), antisymmetric-vanishes, Hermitian-phase preservation, and idempotency. Full suite: 1180 passed under both the gating unittest runner and pytest.

Refs #108.


k-yoshimi:
The first review round verified both delegations raw-byte-identical on adversarial inputs (complex, NaN/Inf, signed zero, wrapped keys, orbital counts to four), derived the FFT sign convention from both production implementations, and MEASURED the two-representation design rationale: on a production-like 64x64 four-orbital table the algebraically identical pipelines differ by ~3e-16 relative -- float summation order, exactly as documented, so neither form should delegate through transforms. Its findings are fixed in 0cc1e5b: the UHFk adjudication note the move had dropped is restored verbatim with references corrected, the delegation import is at module scope (no cycle exists), and the recommended tests are in -- the convention is now anchored to the production transforms rather than a test-local FFT choice, delegation spies prove all three consumer routes reach the shared implementations, and the IEEE-special parity is committed. Full suite: 1185 passed.

k-yoshimi:
Round 2 converged with no findings: the production-anchored convention test fails under both adversarial sign mutations (k-route exponent flip; dense-route FFT reversal), all three consumer routes reach the shared implementation exactly once with inline-regression simulations caught, the restored adjudication note matches the original up to moved-context corrections, and no import cycle exists. Its one optional cleanup (temporary-directory lifetime in the spy fixture) is applied.

k-yoshimi:
A third, deeper round derived the reduction rules independently from the operator definitions (PairHop is the only Hermitian-partner type; PairLift's endpoint reversal is the same operator; the rule set is closed) and found the drift class still alive on one route: the one-orbital simple/load vertex formulation read the raw tables, so a one-sided off-site declaration produced a vertex differing by a measured 0.7 from every other consumer's reading of the same Hamiltonian. Fixed in 2e6bede -- that path symmetrises first (idempotent, so symmetric input is unchanged), pinned through the production dispatch; the physics dispatch now rejects unknown types instead of silently applying the same-operator rule; and the UHFk note no longer claims the documented format is real-only. The committed tests also include dense idempotency for both rules and the ring-vs-S/C asymmetric-declaration agreement the round ran as review evidence. Full suite: 1190 passed.

k-yoshimi:
A fifth, deeper round measured that the round-3 drift fix itself broke a finer invariant: saved artifacts for ALREADY-symmetric configurations changed at the 1e-18 level (256 of 1024 serialized gap values at output precision), because idempotency only holds algebraically after the exponential transform -- the +R and -R exponentials are rounded independently. Fixed in f195ddf: the closure decision moves to the RAW declaration tables, where it is exact (both-ends declarations carry identical literals; 0.5*(x+x) == x in IEEE); closed input now takes the byte-for-byte pre-fix pipeline while one-sided and Hermitian-complex declarations keep the drift fix. The EN/JA simple-mode docs now state the symmetrised reading and the behavioral change, per the round's request. Closure detection and the closed-input bitwise pin are committed. Full suite: 1191 passed.

k-yoshimi:
Round 6 found an extreme-value hole in the round-5 closure detector: it formed the mean before comparing, so an exactly-closed finite coefficient above half of float64's maximum overflowed into a false 'open' -- silently losing the promised raw pipeline, with warnings. Fixed in 1ab3d7d by comparing the table directly with its reversed/transposed partner (no arithmetic), keeping explicit non-finite rejection. The round's requested captures are committed: the huge-coefficient case under warnings-as-errors, and an end-to-end run proving the production flow takes the simple path with the detector firing True and the symmetriser never invoked. A useful cross-check surfaced while writing it: since the auto-selector fix in the previous PR, CoulombInter routes the calc path to the general scheme, so the simple path sees inter-site V only through loaded susceptibilities -- the fixture and comment document that. Full suite: 1192 passed.

PR #130, opened 2026-08-01
Read-time declaration validation, and the Ising coupling read consistently across solvers

k-yoshimi:
## Summary

Two maintainer-decided policy changes, built on the shared declarations module (#108):

**Issue #93 -- reject unclosed declarations at read time.** An interaction file whose declaration pair disagrees beyond a scale-aware tolerance -- including a declared entry with an absent partner -- is a typo, not a different model, and every solver used to do something different with it. Both readers now reject it with a ValueError naming the file, the entry and both values, using the bond resolver's Hermitian-closure rule (X_ab(R) = conj(X_ba(-R))). CoulombIntra additionally rejects off-site / inter-orbital entries (UHFk used to drop them silently). Existing tests that fed such input through files become rejection pins; the mean/zero adjudications stay pinned at the builder level.

**Issue #106 -- UHFk's Ising factor.** The file spec writes the Ising Hamiltonian as J (n_up-n_down)(n_up-n_down) and the RPA/FLEX vertex content was ED-adjudicated against exactly that operator; UHFk alone applied an extra 1/4. The factor is removed, the Ising reference fixture is regenerated, and an equivalence pin proves the change is precisely the factor (quarter-scaled coupling reproduces the historical energies to SCF tolerance). EN/JA file-spec notes document both behavioral changes.

Full suite: 1198 passed, both runners.

Refs #93, #106.


k-yoshimi:
The first review round found the validator failing open in three ways (an absent partner below the absolute tolerance; NaN/Inf pairs whose comparison is false; a complex or non-finite CoulombIntra diagonal) and a set of research-facing documentation contradictions -- most importantly that the UHFk spec page itself still defined Ising as J S^z S^z with the 1/4 built in, the very convention this change removes. All fixed in fcdafd3: membership-first validation with explicit non-finite rejection and a real-diagonal rule; the spec equation now states the shared normalization with the history in a note; the RPA algorithm pages drop the obsolete 1/4 and cancelling-accepted statements; the broken raw directives are gone; and the validation boundary (UHFr's separate reader) is documented. Full suite: 1198 passed, both runners.

k-yoshimi:
Round 2 verified all round-1 fixes by rerunning the five fail-open reproductions and the equivalence-pin mutation, and found the documentation trail deeper than one page: the RPA input specifications themselves still defined Ising in the removed convention (four files), the migration note contradicted itself, and the new test classes sat below unittest.main(). All fixed in 44dc41f -- the four spec pages state the shared density-difference form, the history is an unambiguous three-part statement, the closure note names the exact validated scope including CoulombIntra's rule, and the classes run under direct execution. Full suite: 1198 passed, both runners.

k-yoshimi:
Round 3's full-tree sweep found four more stale sites, fixed in 2a24d65: the UHFk input-index pages still carried the removed S^z definition, the Eliashberg algorithm pages still described the one-sided cos(qR) reading that read-time validation has made unreachable, the migration note overclaimed scope (UHFr's separate real-space reader keeps its own convention -- now stated), and the UHFk equivalence test sat below the main guard. The fail-open and equivalence-pin mutations were re-verified against the fixed code.

k-yoshimi:
Round 4's exhaustive sweep surfaced the final layer, fixed in c1dab98: the bit-for-bit claim is scoped to real closed declarations (the review reproduced a Hermitian-closed complex pair folding to its real part), the JA spec equations and both localization catalogs drop the removed convention, the unspecified-elements statements now mention the missing-partner rejection, the internal flex/rpa comments distinguish reader-bypassing tables from file input, and the conflict-test fixture no longer writes into a committed directory. Full suite: 1198 passed, both runners.

k-yoshimi:
Round 5 measured UHFr's actual Ising reading ([-0.25, +0.25] for J=1, i.e. the S^z convention) against its own spec pages, which documented the unscaled form -- fixed in 7f2afac by stating the convention UHFr actually implements with a cross-reference, and scoping the shared-normalization claims to the k-space solvers. The sweep's remaining current-tense statements describing unclosed declarations as valid are reworded to distinguish reader-bypassing internal tables from file input. Full suite: 1198 passed, both runners.

k-yoshimi:
Final wording pass (acad053): the block-matrix Ising test docstring now names the density-difference form for the unscaled spin table, and the last two comments describing unclosed declarations as accepted are scoped to internal tables built after the read. Full suite green on both runners.

PR #131, opened 2026-08-01
Matsubara tail correction for the Eliashberg pair bubble (issue #86)

### LINKED ISSUE #86: _calc_g2 has no Matsubara tail correction: small-Nmat multi-orbital runs get ~1.5% spurious imaginary eigenvalues
## Summary

`sc._calc_g2` (the pair weight `W = (1/β) Σ_n G(k,iω_n) G(−k,−iω_n)`, shared with `bond_channels._g2_from_green`) performs a **bare truncated Matsubara sum with no tail correction**, unlike `chi0q`, which has `coeff_tail`.

At small `Nmat` this is numerically visible: the truncation makes `W` slightly indefinite, which injects spurious imaginary parts into the Eliashberg eigenvalues of multi-orbital models.

Measured at `Nmat = 64`, `β = 10`, 2-orbital:

- `W` minimum eigenvalue: **−1.3e-2** (should be ≥ 0)
- imaginary fraction of the reported eigenvalues: **~1.5%**
- shift in the leading λ: **~1.5%**

It converges away by `Nmat ≈ 512`.

## Why it matters

`W` being positive semi-definite is what makes the static Eliashberg kernel `K = −ΓW` similar to the Hermitian `K̃ = −√W Γ √W`, hence real-spectrum. When truncation makes `W` indefinite, users see complex eigenvalues and may reasonably mistake them for a broken symmetry or a bug in the vertex, when in fact it is only Matsubara truncation error.

(For the record, the kernel's Hermitian structure itself is intact — `W` Hermiticity follows from `G(k,iω)† = G(k,−iω)` plus the symmetric Matsubara grid, independent of spin structure; measured residuals ≤1.3e-16 including in spin-orbital models. The issue here is positivity under truncation, not Hermiticity.)

## Suggested fix

Either add a tail correction analogous to `chi0q`'s `coeff_tail`, or document the behaviour — e.g. warn when the computed `W` has a significantly negative eigenvalue, pointing at `Nmat` — so that complex eigenvalues from a small-`Nmat` multi-orbital run are read as truncation error.

Found while working on #82; pre-existing, not caused by that PR.

k-yoshimi:
## Summary

Closes #86.

`sc._calc_g2` computed the pair weight `W = (1/beta) sum_n G(k,iwn) G(-k,-iwn)` as a bare truncated Matsubara sum, with no counterpart of `chi0q`'s `coeff_tail`. At small `Nmat` the truncation makes `W` slightly indefinite, which injects spurious imaginary parts into the static Eliashberg eigenvalues of multi-orbital models (measured in the issue: min eigenvalue -1.3e-2 and ~1.5% imaginary fraction at `Nmat = 64`, `beta = 10`).

## The correction

The summand's exact high-frequency tail is `delta_ij delta_lm / wn^2`: the `1/(iwn)` coefficient of `G` is the identity by completeness of the eigenbasis, which holds for dressed (FLEX) Green functions as well, so unlike `coeff_tail` there is no free coefficient. The solver now subtracts that model inside the frequency window and adds its exact full sum `(1/beta) sum_Z 1/wn^2 = beta/4` analytically -- a positive multiple of the identity on the gap space.

Measured on a 2-orbital test model (`beta = 10`):

| Nmat | max error bare | max error corrected | min eig of W bare -> corrected |
|---|---|---|---|
| 64 | 1.6e-2 | 1.6e-3 | -1.6e-2 -> +1.0e-4 |
| 512 | 2.0e-3 | 3.6e-6 | -2.0e-3 -> +5.6e-7 |
| 2048 | 4.9e-4 | 5.7e-8 | -4.9e-4 -> +1.4e-8 |

(The exact band-basis value has min eigenvalue 0 up to rounding.) Positive semi-definiteness of `W` is what makes the static kernel similar to a Hermitian matrix, hence real-spectrum; restoring it removes the spurious complex eigenvalues at their source.

## Behavior and compatibility

- The correction is **on by default**. `[eliashberg] g2_tail = false` (read case-insensitively) reproduces pre-correction results; documented in EN/JA with a migration note.
- When the computed `W` still has a significantly negative eigenvalue, the solver warns and points at `Nmat` (and at `g2_tail` if disabled), so truncation error is not misread as a broken symmetry. The check is skipped with a log line when `norb^4 * nvol` is large enough to rival the solve.
- The dynamic (frequency-resolved) solver never takes this frequency sum and is unchanged; its summand pin now compares against `tail=False` explicitly.

## Tests

New `tests/test_g2_tail.py`:
- corrected sum approaches an **independent band-basis evaluation** of the exact Matsubara sum (analytic pair bubble `(1 - f(ea) - f(eb))/(ea + eb)`), with the bare error >1e-2 and the corrected error at least 5x smaller;
- positivity restoration (bare min eigenvalue < -1e-3, corrected > -1e-10);
- the implemented term is exactly the analytic identity shift `c * delta_ij delta_lm`, uniform in k;
- config plumbing through the public `calc_eliashberg` entry, including a differently-cased key (`G2_Tail`) behaving identically;
- the indefiniteness warning fires on the bare bubble and stays silent on the corrected one.

Full suite: 1203 passed on both runners (pytest and unittest discovery).


k-yoshimi:
Review round 1 (2124e46f): odd or non-positive frequency axes are now rejected by the tail correction with an actionable error (the calc route was already guarded by the RPA layer; loaded Green functions were not), the indefiniteness diagnostic reports significant non-Hermiticity separately instead of silently Hermitizing, states truncation as the likely rather than established cause, and its skip guard bounds actual eigensolver work and memory. PSD-restoration wording scoped to 'in practice'. Added regressions for each branch, a frozen-reference bit-identity pin for the off-switch, a string-boolean config value, and a dressed Dyson-Green convergence case. Full suite green on both runners.

k-yoshimi:
Review round 2 (f704a481): the diagnostic's memory budget now counts the three full-size temporaries it actually allocates (measured ~1.5x the previous two-copy estimate), remaining categorical claims about indefiniteness and complex eigenvalues are stated conditionally in code and EN/JA docs, the size-guard test exercises the byte bound independently, and a new regression pins the previously-unvalidated route: an odd-frequency-grid green.npz accepted by the FLEX loader is rejected by the tail correction with the actionable error. Full suite green on both runners.

k-yoshimi:
Review round 3 (deep review, ea8204b2) found and fixed four items: (1) a temperature-sweep resume started before this change could silently mix pre- and post-correction eigenvalues in one ladder -- the manifest version is bumped so old manifests are refused, and the fingerprint now carries the resolved switch; (2) the correction is asymptotic and can overshoot when the largest retained frequency is below the model's energy scales (measured counterexample at Nmat=2) -- a new window-edge diagnostic warns in that regime and also catches loaded Green functions with a non-identity leading coefficient, and the docs state the premise and the O(Nmat^-3) remainder; (3) the switch now rejects unrecognized values instead of reading misspellings as false; (4) FLEX green files gain beta provenance, validated at load, with restricted frequency axes rejected. The exact-reference fixture gains the stable degenerate-channel branch. Full suite green on both runners.

k-yoshimi:
Review round 4 (d3488f0f): the beta provenance check now fails closed (one finite positive scalar required; NaN/Inf/vector/empty values were previously accepted or crashed incidentally) with a symmetric relative tolerance that also catches high-temperature mismatches, and the window-edge diagnostic handles an empty frequency axis neutrally so the documented grid error is what users see. Added regressions for each case plus a writer-level pin that green.npz stores scalar beta = 1/T. Full suite green on both runners.

k-yoshimi:
Review round 5 (68349893): the beta metadata gate is now typed (real-numeric dtype required before conversion -- complex values previously matched via silent imaginary-part discard, booleans read as 0/1, strings crashed with the wrong exception), the tolerance is explicitly symmetric and pinned at the boundary in both orderings, and a temperature sweep that never runs the Eliashberg solver no longer fails on a malformed value of the solver's own switch. Full suite green on both runners.

k-yoshimi:
Review round 6 (deep review, 3087f911): closed two fail-open holes -- an invalid run temperature (NaN/Inf/boolean) previously slipped through the provenance comparison and fed beta = NaN to the whole Matsubara grid, and a finite but astronomically large beta silently overflowed the pair-bubble accumulation to NaN (now gated, with the tail coefficient computed in dimensionless form and non-finite bubbles rejected outright). The even-grid check now runs before the expensive multiplication, the FLEX loader rejects odd frequency axes at the file boundary, reduced-precision input is promoted, sweep fingerprints are canonical across equivalent spellings of the switch, and dynamic-frequency sweeps no longer parse a switch they never use. A mutation audit added exact pins for every diagnostic threshold, both tolerance sides, frequency-axis content checks, and the manifest schema version. Full suite green on both runners.

k-yoshimi:
Review round 7 (254c6c1c): a subnormal temperature passed the new gate while its reciprocal overflowed to infinity, silently recreating the fail-open comparison -- the helper now validates the resulting beta as finite and positive, the dynamic solver shares the same validated conversion, and the numeric-limit error text distinguishes values one ULP apart. Full suite green on both runners.

k-yoshimi:
Review round 9 (8df58968): the scalar gates now validate the value AFTER narrowing to float64 -- on x86-64 Linux a finite extended-precision value beyond binary64 range previously passed the elementwise check and narrowed to infinity, silently re-opening the provenance comparison, and a sub-range temperature produced the wrong exception type; one shared helper closes all three gates identically. The pair-bubble finiteness scan runs in bounded chunks instead of allocating a tensor-sized mask, the config tests delegate redundant spellings to the shared parser unit tests (cutting dozens of repeated public solves per CI run), and the docs note that leading-eigenvalue crossings under the correction are corrected physics rather than a defect. Full suite green on both runners.

k-yoshimi:
Review round 10 (681bb68e): the chunked finiteness scan flattened with a full copy on the production memory layout -- worse than the mask it replaced; it now flattens in memory order sharing storage, with the layout premise and exact chunk sizes pinned by a test large enough that a whole-tensor scan cannot hide in one chunk. The shared scalar gate rejects single-element containers, and the extended-precision underflow regression's skip condition was computed in the wrong precision (it never ran anywhere) and is fixed to run on the CI platform. Full suite green on both runners.

k-yoshimi:
Review round 11 (96c610a8): the scalar gate now requires a genuine zero-dimensional value (single-element arrays were still accepted through the size check), and the chunk-scan regression asserts the scanned arrays share the pair bubble's storage rather than inferring it indirectly. Full suite green on both runners.

Issue #83, opened 2026-07-26
hwave_sc silently accepts enable_spin_orbital and produces inconsistent results


k-yoshimi:
## Summary

`hwave_sc` (the Eliashberg solver) neither supports nor rejects `[mode] enable_spin_orbital = true`. An SO-mode configuration **runs to completion and prints eigenvalues with no warning**, while several index/orbital-count conventions are inconsistent internally — i.e. it is a silent wrong result.

Reproduced end to end with a 2-physical-orbital × 2-spin model (complex on-site spin-flip, `chi0q_mode="calc"`): the run finished normally and reported eigenvalues `3.123, 2.676, −2.677, −6.218` (Im ~1e-13). No guard, no warning.

## Evidence of the internal inconsistencies

1. **`norb` is the spin-orbital count, never halved to the physical orbital count.**
   `sc.py:4082` (and `eliashberg_dynamic.py:1179`) take `norb = geom_info["norb"]` with no `_so_physical_norb` equivalent.

2. **chi0q is in the spin-block convention but the vertex uses interleaved indices.**
   `sc._build_hamiltonian_k` (`sc.py:1048-1053`) uses transfer indices as-is, i.e. **interleaved** `2*orb+spin`. But `RPA._make_ham_trans` (`rpa.py:350-373`) explicitly remaps interleaved → **spin-block** `spin*norb+orb`, and `validate_chi0q_index_convention` (`rpa.py:32-73`) *requires* the stored chi0q to be tagged `spin_block`.
   Measured: the chi0q returned by `sc._calc_chi0q_internal` vs the same bubble rebuilt from `sc`'s own Green function differs by **27%**; applying the permutation `P = [0,2,1,3]` makes them agree to **8e-15**. (The permutation is the identity only for `norb_phys == 1`, which is why single-orbital SO has not shown this.)

3. **Interaction files are read at the spin-orbital dimension.**
   In SO mode UHFk/RPA read interaction files in *physical* orbital indices, but `sc._build_interaction_k` (`sc.py:1086-1092`) builds them at `norb = 2·N_phys`. In the test fixture, `CoulombIntra` with `U = 2` on physical orbitals 0 and 1 lands as `diag(U) = [2, 2, 0, 0]` over the four spin-orbitals — i.e. `U` is applied to (orb0,↑) and (orb0,↓) as if they were two different orbitals, and physical orbital 1 receives no `U` at all.

[...]

PR #132, opened 2026-08-01
Reject enable_spin_orbital in the Eliashberg solver (issue #83)

k-yoshimi:
## Summary

Implements the short-term resolution of #83: `hwave_sc` now **refuses** `[mode] enable_spin_orbital = true` with an explicit error on both public entries (`calc_eliashberg` and `eliashberg_dynamic.solve_dynamic`), instead of running to completion and printing eigenvalues.

## Why rejection

The issue documents four internal inconsistencies that make a spin-orbital run a silently wrong result rather than an approximation:

1. `norb` is the spin-orbital count, never halved to the physical orbital count.
2. The internally computed chi0q is in interleaved index order while the consumers expect spin-block order (measured 27% discrepancy, exactly the `[0,2,1,3]` permutation).
3. Interaction files are read at the spin-orbital dimension, so `U` lands on (orb0,up)/(orb0,down) as two different 'orbitals' and other physical orbitals receive none.
4. The paramagnetic orbital-space S/C (Kanamori) rules are applied to spin-orbital indices.

Rejecting up front matches the repository's existing no-silent-wrong-result policy (chi0q convention validation, the #93 reader rejection, the #105 off-site hard error). Full spin-orbital support -- physical/spin-orbital `norb` bookkeeping, one agreed index convention, physical-orbital interaction reading, and a spin-generalized vertex -- remains future work, tracked in #83's longer-term section.

## Details

- The key is looked up case-insensitively and coerced with the shared boolean reader, so string forms of false do not trip the guard.
- Tests: truthy forms (bool, string `"true"`, differently-cased key) rejected; falsy/absent forms pass; a `python -O` subprocess regression; both public entries, including a pin that the static path rejects before writing any output file.
- Docs (EN/JA): the FLEX-vs-Eliashberg capability table and the spin-structure limitation section now state the up-front rejection (previously they described only the FLEX-load shape error, while the internal-chi0q path ran silently).

Full suite: green on both runners (pytest and unittest discovery).


k-yoshimi:
Review round 1 (3e99acb2): the flag is now resolved through one strict parser for every consumer -- a string 'false' previously reached the chi0q convention check and RPA as a truthy raw value (with internally inconsistent handling there), and a misspelling silently produced a non-SO run; unrecognized values raise everywhere. hwave_tsweep rejects the configuration in preflight, before any FLEX rung runs, while FLEX-only sweeps remain allowed. Docs scope the limitation to the spin-orbital representation as such and mark the loader-level shape diagnostic as unreachable from the guarded entries. Full suite green on both runners.

Issue #134, opened 2026-08-02
RPA coeff_tail slows chi0q convergence instead of accelerating it: missing G(0-) discontinuity on the reversed equal-time Green factor


k-yoshimi:
## Summary

With `[mode.param] coeff_tail = 1.0` the general-scheme `chi0q` converges to the exact result **~5x more slowly** than with `coeff_tail = 0` -- the opposite of the option's documented purpose. Measured against an independently constructed exact Lindhard function (4-orbital model, T = 0.5, fixed mu; identical numbers with and without `enable_spin_orbital`):

| Nmat | tail = 0 | tail = 1 | penalty |
|---:|---:|---:|---:|
| 256 | 7.916e-4 | 3.938e-3 | 4.98x |
| 1024 | 1.979e-4 | 9.859e-4 | 4.98x |
| 4096 | 4.947e-5 | 2.466e-4 | 4.98x |

Both settings converge at $O(1/N_{\rm mat})$; a correct leading-tail treatment should improve the order, not just the prefactor.

## Root cause (verified with a scalar analytic control)

The frequency-space subtraction $G(i\omega) - a/i\omega$ (`src/hwave/solver/rpa.py:2450`) and the imaginary-time add-back of $-a/2$ (`rpa.py:2553-2556`) are correct **for the forward Green factor**: they reconstruct $G(0^+)$. But the bubble's reversed factor needs the other side of the equal-time discontinuity,

$$G(0^-) = G(0^+) + \beta I,$$

and the time reversal (`rpa.py:2564`) leaves index 0 fixed while `sgn[0] = 1` (`rpa.py:2566`) does not restore the jump. The tail-corrected bubble therefore carries an $O(1/N_{\rm mat})$ error from the single tau = 0 point that is larger than the error the correction removes.

A scalar control confirms the mechanism: adding the missing $+\beta I$ to the reversed zero-time factor alone turns the scalar kernel from $O(N^{-1})$ into $O(N^{-2})$ (measured doubling ratios ~4). For the full rank-4 tensor the endpoint fix gives 4.93e-4 / 1.23e-4 / 3.09e-5 at Nmat = 256/1024/4096 -- better, but still $O(N^{-1})$ from cross-tail terms in off-diagonal components; a fully analytic static bubble-tail correction (subtract the model inside the window, add its exact sum) could reach $O(N^{-3})$ on the symmetric grid.

## Suggested resolution

[...]

PR #135, opened 2026-08-02
Align every RPA Fourier transform to the documented e^{+ikR} convention (issue #133)

### LINKED ISSUE #133: RPA: spin-orbital and non-spin-orbital transfer paths use opposite Fourier sign conventions, flipping the q axis of chi0q between modes
## Summary

The RPA solver builds `H(k)` from the Transfer table with **opposite momentum sign conventions** depending on `enable_spin_orbital`:

- SO path: `ifftn(...) * nvol`, i.e. $e^{+ikR}$ (`src/hwave/solver/rpa.py:488`)
- non-SO path: `fftn(...)`, i.e. $e^{-ikR}$ (`src/hwave/solver/rpa.py:504`)

Both modes then run the same general bubble kernel (`rpa.py:2581-2589`), so the physics per mode is internally consistent, but the **q axis of the stored `chi0q` is flipped between the two modes**. For a model without inversion symmetry the q labels of an SO run and a non-SO run of the same physical system disagree.

## Evidence (verified by direct execution)

A 4x1x1 model with norb_phys = 2 (nd = 4 spin-orbitals) and complex Hermitian hoppings including an on-site spin flip (`spin_mode = 'spinful'`) was run through both paths (the same 4x4 hopping matrix read once as spin-orbitals, once as 4 orbitals). Against an independently constructed exact Lindhard function:

- the SO chi0q matches the reference at the identity index arrangement (with the exact pair-Hermiticity tie $B^*_{acbd}(q) = B_{bdac}(q)$),
- the non-SO chi0q matches only after **reversing q** (using $B^*_{acbd}(-q) = B_{cadb}(q)$),
- after basis relabeling plus $q \to -q$, the two modes agree to `3.89e-16`.

So the discrepancy is exactly one momentum-sign convention, not a layout or physics difference.

## Why it matters

- chi0q/chiq files produced in SO vs non-SO mode carry incompatible q labels for non-centrosymmetric models; any consumer comparing or combining them (plots, downstream tools, regression baselines) silently mixes q and -q.
- The documented Wannier90-style convention is $H(k) = \sum_R e^{+ik\cdot R} H(R)$; whichever internal path disagrees with the documentation is mislabeling its output axis, and the sign choice should be adjudicated against the documented convention and against UHFk before changing anything.

## Suggested resolution

1. Adjudicate which sign matches the documented Transfer-file convention (docs + UHFk's construction).
2. Align both RPA paths to it (or, if the long-standing non-SO behavior must be preserved for compatibility, document the SO path's flip and remap at the boundary), with a regression pinning chi0q(q) of a non-centrosymmetric model against an analytic reference in both modes.

Found during the exact-diagonalization validation campaign for spin-orbital support (see #83 discussion for the roadmap); the spinful chi0q itself was validated against the exact free-fermion result in the course of the same work.

k-yoshimi:
## Summary

Closes #133 (the coeff_tail finding is #134, separate).

The documented convention -- $c_i = N^{-1/2} \sum_k e^{+ik r} c_k$, hence $\varepsilon(k) = \sum_R t(R) e^{+ikR}$ and $W(q) = \sum_R W(R) e^{+iqR}$, the Wannier90-style sign that UHFk already follows -- was used by the RPA spin-orbital transfer branch only. The non-spin-orbital transfer, Extern, trans_mod, and $H_0$ rebuild paths, and the interaction/Fierz FFTs (shared by BOTH modes), used $e^{-ikR}$.

**Consequences (each verified by direct execution):**
- The non-spin-orbital mode was a self-consistent global $k \to -k$ relabeling: stored chi0q/chiq of non-centrosymmetric models carried negated momentum labels, and a UHFk-produced `green_init` entered with mismatched labels.
- The spin-orbital mode mixed the two signs internally and solved $\chi_q = [1 + \chi_0(q) W(-q)]^{-1} \chi_0(q)$. This is invisible whenever $W(q) = W(-q)$ -- true for every committed fixture -- but wrong for directional inter-orbital bonds, which are legal Hermitian-closed input.

## Change

All six R->k builds in rpa.py now use `ifftn * nvol` ($e^{+ikR}$), and hwave_sc's two k-space builders follow (their contract is 'the same convention as the solver core'). The chi0/chi convolution machinery is label-agnostic and untouched. FLEX inherits the corrected `Interaction` class, so its vertex and transfer enter with one sign as well.

## Tests

New `tests/test_rpa_fourier_sign.py`, red-verified against the pre-fix code (3 of 4 fail there, exactly the sign-sensitive ones):
- $\varepsilon(k)$ pinned to the documented analytic form in both modes, with complex non-centrosymmetric hoppings;
- the assembled vertex's density slots pinned to $V_{ab}(q) = \sum_R X_{ab}(R) e^{+iqR}$ using a **directional inter-orbital bond** (a same-orbital complex pair is sign-blind: the closure symmetrisation folds its phase to a real coefficient);
- the spin-orbital chiq pinned against an **analytic-W** same-q assembly, with non-vacuity assertions on both premises (q-asymmetric $\chi_0$; $W(q) \neq W(-q)$); the pre-fix spin-orbital code matches the $W(-q)$ assembly instead.

Existing convention pins updated to the documented sign (sc builder phase pins, offsite-Vq phase pins, dense-route anchor, the orientation test's production comparison, and the two-orbital analytic reference). Full suite: 1271 passed on both runners.

## Migration

Momentum-resolved outputs (chi0q/chiq) of **non-centrosymmetric** models produced before this fix carry globally negated q labels relative to current output; centrosymmetric models are unaffected. Documented in the EN/JA algorithm pages.


k-yoshimi:
Review round 1 (80c4e5e0): every momentum-space NPZ now carries a machine-readable Fourier-sign marker validated at all five loader families, with unmarked legacy files accepted only when their content is elementwise q-even (the exact condition under which both conventions coincide); the sweep manifest version is bumped so pre-fix rungs cannot mix into a resumed ladder; the migration note states the precise invisibility condition and that pre-fix spin-orbital results with directional bonds are wrong rather than relabeled; stale sign comments fixed; gate and Extern-sign regressions added. Full suite green on both runners.

k-yoshimi:
Review round 2 (4fa9ed11): closed three bypasses found by adversarial probing -- the q-axis size-search heuristic (frequency-axis collision at nfreq == nvol, and expanded reference layouts skipping the gate) is replaced by structural layout resolution with three-axis support; the evenness tolerance is now pair-local so a small channel cannot hide its own order-one asymmetry beneath a large channel's amplitude, with non-finite payloads rejected; and the dynamic gap writer/seed-loader pair (which chains between temperature-sweep rungs) is stamped and gated. UHFk green outputs now carry the stamp too, and the shared constant replaces string literals. Full suite green on both runners.

k-yoshimi:
Review round 3 (535c9c4c): the 6D layout resolver now matches the full raw and reference patterns (a reference file with norb equal to the lattice volume was previously misrouted onto its orbital axis and bypassed the gate -- reproduced and pinned), genuinely ambiguous shapes fail closed unless the marker decides, and the evenness comparison is normalized before differencing so finite values near the float64 maximum cannot overflow both sides to infinity and slip through. Non-cubic, permuted-axis, negative-axis, zero-vs-denormal, and roundoff-asymmetry cases are pinned, and the dynamic seed loader diagnoses non-6D files explicitly. Full suite green on both runners.

k-yoshimi:
Review round 4 (fc5c7748): IR-native susceptibility and Green files are now gated like uniform ones (the earlier exemption rested on a claim the review disproved by execution), the 6D layout resolver accepts only its two complete patterns and fails closed otherwise (the partial fallback had re-opened the misroute for mismatched-grid reference shapes), the normalization scale is clamped so denormal-scale payloads are not spuriously rejected, and the rejection diagnostic no longer overflows for near-maximum payloads. Unit-level callers without a configured lattice skip the gate rather than checking against a fabricated grid. Full suite green on both runners.

k-yoshimi:
Review round 5 (64d2d40f): unknown 6D layouts are now rejected in the loader independent of provenance -- a matching marker previously short-circuited the check and the converter silently reshaped the file, reinterpreting orbital axes as data -- and the marker itself must be exactly one value (multi-element arrays authorized via their first element; empty ones crashed incidentally). Both pinned. Full suite green on both runners.

k-yoshimi:
Review round 6 (a5f9216d): the chi0q loader now receives the orbital count from the production entry and validates the exact supported layouts before any provenance handling (malformed shapes previously slipped through, and the raw/reference ambiguity disappears once norb is known), and the evenness scale is computed from real and imaginary components so finite complex values whose magnitude overflows cannot erase their own asymmetry. Both pinned with marked and unmarked cases. Full suite green on both runners.

k-yoshimi:
Review round 7 (7e5d7f11): the spin-diagonal layouts are validated over their full trailing shapes (partial slices accepted malformed arrays), and the layout matched by the orbital-count gate now directly selects the momentum axes -- the norb-blind routing previously re-ran afterwards and rejected shapes the gate had already disambiguated. Both reproduced cases pinned, with the ambiguous shape loading correctly under each orbital count. Full suite green on both runners.

k-yoshimi:
Review round 9 (deep, 99b0122d): the legacy-acceptance scans are rewritten to be memory-bounded (the previous pipeline peaked at ~5.5x the payload, which would have made large legacy files unusable; now ~30 MB regardless of size, with a tracemalloc pin), the finiteness check runs for marked files too, RPA's green_init reader now validates a present marker on every branch (a file explicitly recording the wrong convention was accepted before), UHFk's eigen output is stamped, the FLEX tutorial's remaining old-sign formula and the sample plotting script's swapped meshgrid axes are fixed, and the output-file specification documents the marker. The review also verified three cross-solver physics chains and the stored wavevector labels numerically. Full suite green on both runners.

k-yoshimi:
Review round 10 (d7eeb3cd): the legacy scans chunk on two levels so even a single momentum plane larger than the element bound stays memory-bounded (reversed indices are computed per chunk, no full-size index grids -- pinned by tracemalloc regressions on exactly the layouts the review measured at 4.5-6.6x payload), signed-integer payloads are cast per block so INT64_MIN cannot zero the scale via overflow, and marker validation now runs unconditionally at every loader so a file recording the wrong convention is rejected even when the layout cannot be identified. The chunked semantics themselves were verified equivalent to the full-tensor check on 84 fixtures in review. Full suite green on both runners.

PR #201, opened 2026-09-18
Bond-resolved dynamic Eliashberg pairing kernel (in-process and hwave_sc entries)

k-yoshimi:
## Summary

Bond-resolved, frequency-resolved (dynamic) Eliashberg pairing kernel built from the FLEX bond gate's own vertices and dressed susceptibilities. It closes the gap that `frequency = "dynamic"` with `bond_channels = true` used to refuse, and it is what the Esirgen–Schüttler–Bickers T_c(V₁) and the Onari-type inter-site pairing comparisons need.

Two entries share one kernel:

- **In-process** (`[mode.param] longitudinal_bond_pairing = "singlet" | "triplet" | "both"`): at the end of the FLEX solve the pairing eigenvalue is computed from the gate's resident `S`, `C` and `chibar`; outputs `eliashberg_bond_<type>.npz`, `gap_bond_<type>.dat`, `eigenvalue_bond_<type>.dat`, written after every FLEX artifact. A pairing failure never loses the FLEX result (recorded as `pairing_<type>_error`).
- **Post-processing** (`hwave_sc` with `frequency = "dynamic"`, `bond_channels = true`): reads the bond archive `longitudinal_bond.npz` (schema 2, which now carries `S_bond`, `C_bond`, `norb`) and `green.npz` of the same FLEX run; the archive members are consumed one at a time.

The pairing vertex is `Γ_s = 1.5 S χ_s S − 0.5 C χ_c C`, `Γ_t = −0.5 S χ_s S − 0.5 C χ_c C` in the bond-major index, plus the frequency-independent bare term from the channel-0 blocks of the same vertices. The kernel acts on a k-resolved gap with the bond form factors realised as real-space rolls (2B spatial transforms per application), on the uniform Matsubara grid or on the bosonic IR basis (`matsubara_basis = "ir"`, with a componentwise fit-residual check and an optional retained constant). Memory admission (`device` / `host` / `stream` residency) is estimated before the FLEX solve for the in-process entry and again per channel.

### Behaviour change on the existing on-site IR path

`IRAxis.u_matsubara_sum` replaces `u_zero_plus` in the instantaneous term of `eliashberg_kernel_ir`: the frequency-independent vertex multiplies the unregularised Matsubara sum, i.e. the midpoint `0.5 (F(0⁺) − F(β⁻))` of the τ = 0 jump, not the one-sided `F(0⁺)`. Measured scope of the change: on a pair amplitude of definite frequency parity the two prescriptions coincide, so eigenvalues of parity-projected solves and the eigenvalue-solver spectrum are unchanged; the old kernel mapped odd-frequency components into the even sector (parity leakage 8e-2, now 2e-14), which made the power iteration's parity guard switch off the projection (a triplet run could then converge to the singlet-sector eigenvalue) and put an even-parity admixture into odd-channel gap functions. The instantaneous vertex is non-zero for every model in the singlet channel (pure Hubbard included: +U) and for every model with an inter-orbital or off-site interaction in the triplet channel. The tutorial (EN/JA) carries a change note saying who should re-run. `u_matsubara_sum` is statistics-dependent (fermionic: difference, bosonic: sum of the endpoint values); only fermionic axes consume it today.

### Validation

- Oracle tests: at B = 1 the uniform kernel reduces to the on-site `eliashberg_kernel_dynamic`; in the flat-frequency limit the instantaneous term equals the exact-diagonalisation-adjudicated static bond kernel (`make_bond_kernel(part="instantaneous")`) at norb = 1 and norb = 2 (diagonal, off-diagonal and full 2×2 `CoulombInter`, both pairing types, 1e-15); a brute-force real-space oracle at norb = 2, B = 3.
- In-process and post-processing eigenvalues agree to 1e-15 on the same run; IR and uniform agree within 1.7 % (singlet) / 0.9 % (triplet) at Nmat = 128 with the expected Nmat⁻² convergence.
- Flag-off runs are byte-identical to the reference revision 65719b74 (all FLEX artifacts; the bond archive differs exactly by the four new members). CI provisions that reference checkout.
- Full suite green locally (Python 3.13): `unittest discover tests/` 2548 tests OK, `pytest tests/ --cov` 2883 passed with 88.8 % coverage; heavy tests registered in `tests/heavy_tests.py`. Sphinx builds of both docs trees show only the pre-existing messages.
- Three rounds of code review and one documentation review; every remaining finding is either fixed on this branch or listed below as a follow-up.

### Docs

`rpa/filespecification/config/index_config.rst`, `rpa/filespecification/outputfiles/chiq_rpa.rst` (archive schema 2), `flex/tutorial/tu-index.rst` (in-process pairing), `rpa/tutorial/sc-index.rst` (post-processing, IR change notice), EN and JA.

### Follow-ups (filed separately)

ARPACK eigenpair selection vs `num_eigenvalues`; the on-site IR retained-constant alignment; an inter-orbital exact-diagonalisation check of the full pairing vertex; a sidecar archive layout; hardening items from the code review; a re-examination of the issue-57 migration note's "pure Hubbard unaffected" statement (the bare singlet term of a pure Hubbard model is +U, not zero; the shipped comparisons did not expose the aliasing because their pair amplitude is even in frequency).

https://claude.ai/code/session_01QRqmKRgf8s2yi9CUV5phcL


k-yoshimi:
Validation of the kernel against the extended-Hubbard T_c(V1) of Esirgen, Schüttler and Bickers (PRL 82, 1217) is recorded in #208: on 32x32 the V1 = 0 and V1 = 0.5 transition temperatures agree with the paper to about 1 %; the V1 = 1.0 point is still running.

